Issue #34, opened 2026-07-09
Icons are not top-aligned


TruthOf42:
As a user, when I put an icon with no status and an icon with a status next to each other the two icons should appear vertically aligned

Icons shown on the floor plan are aligned at the bottom, they should be aligned at the top

<img width="139" height="100" alt="Image" src="https://github.com/user-attachments/assets/aa30e45e-7a5f-46b0-88ab-29655834c251" />

PR #40, opened 2026-07-10
Six item kinds, eight furniture symbols, a handed sectional, and a documented aspect-ratio trap

amadeobriones:
Four independent changes, each with tests. Split into two commits so any part can be dropped.

> **Maintainer note:** description edited before merge to match the final code — the original text described an aspect-ratio *change* that the branch itself later reverted in favour of a comment; see the first section. Test count updated 119 → 126. No code changes made. — @nicosandller

### `preserveAspectRatio` — documented, not changed

The root `<svg>` renders with `preserveAspectRatio="none"`, so whenever the container's aspect ratio differs from the canvas, the whole plan stretches: circles become ellipses. An earlier revision of this branch "fixed" that with `xMidYMid meet` — and that fix was **wrong**: the `.items` overlay is HTML positioned with raw percentages of `.stage`, and it does not letterbox. With `meet`, the moment anything overrides the stage's ratio (card-mod, a grid row count), the drawing shrinks away from the badges and every icon drifts off the wall it was placed on. `"none"` stretches both layers identically: distorted, but aligned.

The final code therefore **keeps `none`** and adds a comment at the `<svg>` explaining the trap, so the next person doesn't repeat the same wrong fix. The real fix — letterboxing the SVG and the overlay together in one aspect-ratio box — is left as a follow-up.

### Six item kinds

`ItemKind` was `light | switch | sensor | binary_sensor | climate | cover | generic`. Everything else fell to `generic` and rendered `mdi:circle`, so a television and a doorbell looked identical.

Adds `media_player`, `fan`, `camera`, `lock`, `humidifier`, `vacuum` to `ItemKind`, `kindFromEntity()` and `defaultIcon()`, with state-aware icons where HA has them (`mdi:television` / `-off`, `mdi:lock` / `-open`).

This also fixes a small bug in `entityDefaultIcon()`: it returned early when the entity had no `device_class`, so a domain that has a perfectly good state-dependent icon never got one. The domain lookup now happens before that guard. One existing test asserted the old behaviour and has been updated.

New `entityIsActive()` teaches the on/off test the domains that never say `on`: a lock is `unlocked`, a vacuum is `cleaning`, a camera is `recording`. Outages are never active. Tap behaviour is untouched — locks still open more-info, they do not blind-toggle.

### Eight furniture types

`washer`, `dryer`, `dishwasher`, `waterHeater`, `airHandler`, `bathtub`, `vanity`, `sectional` — each drawn as a plan symbol rather than a labelled box. `dev/symbols.html` renders all 23 types on one sheet.

### A sectional with a handedness

`hand: "left" | "right"`, default `"right"`. The chaise extends to the right when you face the sofa from the front; the symbol's back is at local `-y` like every other furniture symbol, so `angle` composes normally. `"left"` is the exact mirror.

`sectionalPoints()` is a pure function, so the tests can assert what actually matters: six corners rather than four, area equal to `w*h` minus the notch, the chaise on the correct side of the front edge, and left being right mirrored rather than a differently-shaped sofa.

### Tests

103 → 126. `tsc --noEmit` clean, bundle 164.22 kB.

`FURNITURE_LABELS` is a `Record<FurnitureType, string>`, so the compiler now refuses a new furniture type that has no label. `FURNITURE_DEFAULT_SIZE` was already one. The render switch is not exhaustive in the type system, so there is a test that renders every type.

### Note on conflicts

This touches `src/editor.ts` to add the type dropdown and the conditional "Chaise side" row. @shauneccles's `ha-form` migration (#44) replaces those hand-rolled rows entirely; merging this first and letting #44 absorb the conflict.

### DIFF
--- a/dev/symbols.html
+++ b/dev/symbols.html
@@ -1,0 +1,6 @@
+<!doctype html><html><head><meta charset="utf-8"><title>symbols</title>
+<style>:root{color-scheme:light;--primary-text-color:#212121;--primary-color:#03a9f4;
+--card-background-color:#fff;--divider-color:rgba(0,0,0,.12)}
+body{margin:0;background:#eceff1;padding:10px;font-family:system-ui}
+#host{background:#fff;width:1100px}</style></head>
+<body><div id="host"></div><script type="module" src="./symbols.ts"></script></body></html>
--- a/dev/symbols.ts
+++ b/dev/symbols.ts
@@ -1,0 +1,41 @@
+import "../src/index";
+import type { FloorplanCardConfig } from "../src/types";
+import { FURNITURE_DEFAULT_SIZE } from "../src/types";
+
+if (!customElements.get("ha-card")) {
+  class C extends HTMLElement {
+    set header(v: string | undefined) { this._h = v; this._r(); }
+    private _h?: string;
+    connectedCallback() { this._r(); }
+    private _r() {
+      if (!this.shadowRoot) this.attachShadow({ mode: "open" });
+      this.shadowRoot!.innerHTML = `<style>:host{display:block;background:#fff}</style><slot></slot>`;
+    }
+  }
+  customElements.define("ha-card", C as any);
+}
+if (!customElements.get("ha-icon")) customElements.define("ha-icon", class extends HTMLElement {} as any);
+
+const TYPES = Object.keys(FURNITURE_DEFAULT_SIZE) as Array<keyof typeof FURNITURE_DEFAULT_SIZE>;
+const COLS = 6, CELL = 260;
+const furniture = TYPES.map((t, i) => {
+  const { w, h } = FURNITURE_DEFAULT_SIZE[t];
+  return { id: t, type: t, x: (i % COLS) * CELL + CELL / 2, y: Math.floor(i / COLS) * CELL + CELL / 2, w, h };
+});
+// both hands of the sectional, side by side
+furniture.push({ id: "sectional_left", type: "sectional", hand: "left",
+  x: (TYPES.length % COLS) * CELL + CELL / 2, y: Math.floor(TYPES.length / COLS) * CELL + CELL / 2,
+  w: 230, h: 180 } as any);
+
+const texts = furniture.map((f) => ({ id: "t_" + f.id, x: f.x, y: f.y + CELL / 2 - 26, text: f.id, size: 15 }));
+
+const config = {
+  type: "custom:easy-floorplan-card", title: "Furniture symbols",
+  width: COLS * CELL, height: (Math.ceil((TYPES.length + 1) / COLS)) * CELL,
+  grid: CELL, walls: [], openings: [], items: [], texts, furniture,
+} as unknown as FloorplanCardConfig;
+
+const card = document.createElement("easy-floorplan-card") as any;
+card.hass = { states: {}, formatEntityState: () => "" };
+card.setConfig(config);
+document.getElementById("host")!.appendChild(card);
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -11,6 +11,7 @@ import type {
   FloorText,
   Furniture,
   FurnitureType,
+  SectionalHand,
   ItemKind,
   ItemDisplay,
   Tracker,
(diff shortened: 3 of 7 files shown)


shauneccles:
Heya friend.

Nice work on the furniture, it looks awesome - really impressed!

I see you spotted the ha-form branch!

I've also got a full-size editor cooked up that I lead with - it's got a lot of fixes that ha-forms needed.

I'll be dropping a PR soon and a draft follow on for ha-forms.


I'm keen not to overwhelm @nicosandller and really enjoying working with this project.

How is this velocity for you @nicosandller?

amadeobriones:
Hi! Thank you for the kind words, glad you like the furniture addition

Apologies for the PRs landing all at once. I was moving fast with an AI agent and opened them before I'd reviewed them properly, and I didn't mean to add to the queue you're already juggling.

I'm closing them. The Codex review was right on all counts: the new lock/vacuum/camera icons can never reach their active state (none of those domains ever reports on), showIcon: false regressed, and the preserveAspectRatio change letterboxes the SVG but not the HTML overlay, so badges drift off their walls. All three are fixed on my fork if they're useful to anyone.

I'll keep working there for my own instance. Really nice project, and I hope to enjoy watching it grow.

shauneccles:
Oh no please that's not what I intended! I just wanted to let you know how much I liked the features you added and figured I'd check in with the Author while we're here since it's a new collaboration and I thought it would be nice to ask.

nicosandller:
Heya guys! I really like your excitement and I'm enjoying the ideas and changes you're all proposing. 

@shauneccles full steam ahead ! I can do a few PRs per day I think :).  

I reopened the PR! Hope you don't mind !

shauneccles:
Hey @amadeobriones — in case you missed it above: @nicosandller reopened this PR and said full steam ahead, a few PRs a day works for him. So please don't read my earlier check-in as anything but enthusiasm — it was a brand-new collaboration and I wanted to make sure we weren't flooding anyone. We weren't. 🙂

On the Codex findings: a reviewer — human or AI — finding bugs is the process *working*, not a mark against the PR or its author. All software has bugs; they only matter if they go unaddressed, and you'd already fixed all three on your fork main within hours ([`0f7f970`](https://github.com/amadeobriones/easy-floorplan/commit/0f7f970), [`1b46bf3`](https://github.com/amadeobriones/easy-floorplan/commit/1b46bf3), [`fb6cdbf`](https://github.com/amadeobriones/easy-floorplan/commit/fb6cdbf)). If you push those three onto `feat/item-kinds-and-aspect`, this looks ready to land.

I've also opened #46 — a map linking the open issues here to the fork commits that already fix them, with a suggested order to keep PRs small (my #43/#44 first, since your fork builds on that branch). Zero obligation and entirely your pace. The only goal is that we all feel safe to move fast together, without the value scattering across three repos.

amadeobriones:
Thanks for reopening, and no worries at all on the earlier check-in. I've applied two of the three fixes to `feat/item-kinds-and-aspect` — the `preserveAspectRatio` revert and the lock/vacuum/camera active-state fix — so this should be ready to land.

On the third (the `showIcon: false` fix): that one repairs a regression from the icon-alignment change in #34, which isn't part of this branch — `showIcon: false` already renders correctly here — so I left it out. It belongs with #34's work.

On the collaboration side, I've shared where I'm at over in #46 — worth a look. Thanks for making it easy to jump in.

nicosandller:
Reviewed and merging — thanks @amadeobriones, this is a strong first contribution. Ran the full suite locally (126/126, typecheck + build clean) and rendered `dev/symbols.html`: all eight new symbols read as proper plan glyphs and the two sectional hands are true mirrors.

Two notes:

1. I edited the PR title/description before merging so the squash commit matches the final code — the original text described the `xMidYMid meet` change that the branch itself later reverted. For the record, keeping `"none"` was the right call: the `.items` overlay doesn't letterbox, so `meet` would desync badges from walls exactly in the mismatched-ratio case. The do-not-fix comment is genuinely useful; only suggestion there is to move it from an HTML comment inside the template to a TS comment above it in a follow-up, so the prose doesn't ship in the bundle and the DOM.

2. One small follow-up worth doing (happy to take it): the editor's item overlay still uses the inline `on/open/home/playing` check, so an unlocked lock renders `mdi:lock-open-variant` on the dashboard but `mdi:lock` in the editor preview. Switching the editor to your `entityIsActive()` makes the preview truthful.

Also appreciated: leaving `ACTIVE_DOMAINS` alone so locks/vacuums/cameras open more-info on tap rather than blind-toggling, and the mirrored-polygon property test on `sectionalPoints`.

PR #42, opened 2026-07-10
Honour the entity-registry icon override

amadeobriones:
Found and reported by @shauneccles (shauneccles/easy-floorplan#2), as the first of the two gaps he split out of the display-precision work. This is that half.

### Problem

A user who sets an icon on an entity in **Settings → Entities** sees it everywhere in Home Assistant except on the floorplan.

That override lives at `hass.entities[entityId].icon`. It never reaches `attributes.icon`, which is the only place `_itemIcon` looks — so the card silently diverges from the rest of the dashboard, and there is no way for the user to tell why.

### Fix

`_itemIcon` consults it, directly below the card's own `icon` override:

```
item.icon  →  registry icon  →  attributes.icon  →  device_class  →  kind default
```

That is the order HA's own `entityIcon()` uses. An icon set on the item is the card's config and is more specific than the registry; the registry is the user's decision and is more specific than whatever the integration supplied.

`custom-card-helpers` does not declare `entities` on `HomeAssistant`, though HA has passed it to cards since 2023.4, so it is declared here as optional. When it is absent — as in every existing test — the precedence is exactly what it was.

### Scope

Deliberately minimal. @shauneccles's issue also asks for **state-based** icons (a light that is `mdi:lightbulb` on and `mdi:lightbulb-outline` off); that is the second half and it is in #40, which adds `entityDefaultIcon` state handling along with six new item kinds.

He has also extracted this resolution into a pure `resolveItemIcon` in `render.ts` on his branch, which is the right home for it and would make it unit-testable. I have kept this patch inline so it does not collide with that; happy to rebase on top once it lands.

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -2365,7 +2365,8 @@ export class FloorplanCardEditor extends LitElement {
   private _renderItemOverlay(it: FloorItem, c: FloorplanCardConfig): TemplateResult {
     const selected = this._isSel("item", it.id);
     const st = it.entity ? this.hass?.states[it.entity] : undefined;
-    const icon = resolveItemIcon(it, st);
+    // Pass the registry icon here too, so the editor preview matches the card.
+    const icon = resolveItemIcon(it, st, it.entity ? this.hass?.entities?.[it.entity]?.icon : undefined);
     const label = it.name || it.entity || it.kind;
     const size = it.size ?? DEFAULT_ITEM_SIZE;
     const showIcon = it.showIcon ?? true;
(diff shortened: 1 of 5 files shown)


shauneccles (commented):
Looks good just needs a rebase

amadeobriones:
Rebased — thanks @shauneccles! `resolveItemIcon` became the shared card/editor icon helper in #43, so I moved the registry-icon step onto it (it takes an optional `registryIcon` arg now, since it gets the state object rather than `hass`). Both the card and the editor preview pass it, so they stay in sync. Added a small unit test for the precedence.

Heads-up that #57 also touches `resolveItemIcon` (making `entity` optional) — they're compatible, just a trivial reconcile whichever lands first. Should be clean to land now.

nicosandller (approved):
Looks good to me. Thank you for the work @amadeobriones  :).

PR #44, opened 2026-07-10
Editor: migrate form rows to ha-form/HA selectors; standard tap/hold/double-tap actions on items

shauneccles:
> **Stacked on #43** — the first six commits here are that PR; only the last six are new. Draft until #43 merges, after which this rebases to show just its own diff.

## Problem

The editor's form rows were hand-rolled `<input>`/`<select>` elements: they don't match HA's look/density/RTL, labels weren't programmatically associated, the icon picker was effectively never available (the entities-editor loading trick doesn't define it), and items had a single hard-coded tap behavior.

## Change

**Schema-driven forms.** Every element section (opening, item, text, furniture, tracker geometry, wall coordinates) and the Project panel now renders through one `_renderForm` helper backed by pure, unit-tested schema builders (`editor-forms.ts`). When `ha-form` is defined you get native HA selectors — filtered entity pickers, a working icon picker (ha-form lazy-loads its own selectors), number fields with units and slider/box modes, helper text, and properly associated labels. Outside HA (dev harness, load failure) the same schema renders as plain inputs and upgrades in place via `whenDefined`. `ha-form` is loaded through the **button-card** editor: verified against the frontend source that the entities editor does *not* import ha-form, while hui-button-card-editor does (plus the whole `ui_action` selector chain). Selector shapes are pinned to forms verified compatible from HA 2024.1 through current dev (`entity: { filter: [...] }`, no `flatten`, no select `mode: "box"`).

ha-form's per-keystroke `value-changed` events are diffed against the schema, normalized (empty-optional → unset, invalid required numbers keep the old value, range clamps, angle wrap), and routed through the undo-burst discipline from #43 — typing runs and slider drags stay one undo step each.

**Standard actions on items.** `tap_action` / `hold_action` / `double_tap_action` (ui_action selectors in the editor, canonical gesture directive on the card: 500ms hold, double-tap window, keyboard activation). The executor handles both `perform-action` and the legacy `call-service` spelling — the ui_action selector emits the former on ≥2024.8 and the latter before, and `custom-card-helpers`' `handleAction` silently drops `perform-action`, so the card ships its own (tested). **Unconfigured items behave exactly as before**: toggle for controllable domains, more-info otherwise. Badges are now keyboard-reachable (`role=button`, Enter/Space).

Kept custom deliberately: color swatch+text rows (HA has no CSS-color-string selector; ours accept `var(--x)` / empty-for-theme-default), tracker sensor sub-editors, the wall Length resize control, and the snap control.

## Verification

- typecheck, build, vitest: 157 tests (34 new: schema conditionals, diff/normalize guards, action resolution/execution for both spellings).
- 24-check Playwright pass against the dev harness (fallback path): all #43 checks plus select-commit = one undo step, cleared required number keeps the old value, two separate slider drags = two undo steps.
- The ha-form path itself needs a pass in real HA (the harness can't load HA components) — that's the main thing to poke at while this is in draft: each dropdown/picker, the action editors, and fullscreen + pickers together.

https://claude.ai/code/session_01TMzXLjqWUr9QCeArktZqBw

### DIFF
--- a/dev/symbols.html
+++ b/dev/symbols.html
@@ -1,0 +1,6 @@
+<!doctype html><html><head><meta charset="utf-8"><title>symbols</title>
+<style>:root{color-scheme:light;--primary-text-color:#212121;--primary-color:#03a9f4;
+--card-background-color:#fff;--divider-color:rgba(0,0,0,.12)}
+body{margin:0;background:#eceff1;padding:10px;font-family:system-ui}
+#host{background:#fff;width:1100px}</style></head>
+<body><div id="host"></div><script type="module" src="./symbols.ts"></script></body></html>
--- a/dev/symbols.ts
+++ b/dev/symbols.ts
@@ -1,0 +1,41 @@
+import "../src/index";
+import type { FloorplanCardConfig } from "../src/types";
+import { FURNITURE_DEFAULT_SIZE } from "../src/types";
+
+if (!customElements.get("ha-card")) {
+  class C extends HTMLElement {
+    set header(v: string | undefined) { this._h = v; this._r(); }
+    private _h?: string;
+    connectedCallback() { this._r(); }
+    private _r() {
+      if (!this.shadowRoot) this.attachShadow({ mode: "open" });
+      this.shadowRoot!.innerHTML = `<style>:host{display:block;background:#fff}</style><slot></slot>`;
+    }
+  }
+  customElements.define("ha-card", C as any);
+}
+if (!customElements.get("ha-icon")) customElements.define("ha-icon", class extends HTMLElement {} as any);
+
+const TYPES = Object.keys(FURNITURE_DEFAULT_SIZE) as Array<keyof typeof FURNITURE_DEFAULT_SIZE>;
+const COLS = 6, CELL = 260;
+const furniture = TYPES.map((t, i) => {
+  const { w, h } = FURNITURE_DEFAULT_SIZE[t];
+  return { id: t, type: t, x: (i % COLS) * CELL + CELL / 2, y: Math.floor(i / COLS) * CELL + CELL / 2, w, h };
+});
+// both hands of the sectional, side by side
+furniture.push({ id: "sectional_left", type: "sectional", hand: "left",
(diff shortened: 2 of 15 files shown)


shauneccles:
Completed the manual pass this branch was waiting on, in a real HA instance (core 2026.7.1): the ha-form path renders everywhere (entity/icon pickers, textfields, selects with domain-aware action defaults), dropdowns work inside the fullscreen top layer, entity selection from the picker works, and keyboard input stays contained to the fields. Details and one found-and-fixed Escape-handling bug are written up on #43 (fix commit d7c8d65, merged here as 3887363). From my side both PRs are ready for review.

nicosandller (approved):
Thanks for this. Looks good. I will resolve the merge conflicts and merge.

nicosandller:
Resolved the merge conflicts with main — commit 33047a8, pushed here directly (maintainer edits enabled). Your commits are untouched.

Why this one conflicted so widely: #43 landed on main as a **squash**, so this stack shares no history with it — every #43 hunk conflicted against its own squashed copy (those all resolved to identical content), leaving three real reconciliations:

- **render.ts / floorplan-card.ts / render.test.ts** — same resolution as on #43's branch: #40's domain-aware `entityIsActive` stays the shared on-state predicate (your `resolveItemIcon` now uses it, so lock/vacuum/camera items reach their active icons), single `isEntityOn`, union of test suites.
- **editor.ts** — took this branch's schema-driven version wholesale; the main side was the hand-rolled rows it replaces.
- **editor-forms.ts** — ported #40's editor additions to their new home: the eight furniture types + labels, and a conditional **Chaise side** field in `furnitureForm()` (sectional only), in the same shape `openingForm` uses for hinge/slide. @amadeobriones — this is exactly the move you offered in #40; if the version on your fork differs, feel free to adjust on top.

Verified post-merge: `tsc` clean, **180/180** tests, build clean (181.4 kB); harness pass covering all 23 furniture glyphs in the + Add popover, sectional `hand` editing end-to-end through the (fallback) form, card/editor lock-icon consistency, and default tap → `homeassistant.toggle` through the new action pipeline.

Still open before merge, from the review: the real-HA `ha-form` pass you flagged (each dropdown/picker, action editors, fullscreen + pickers together), and committing the 24-check Playwright suite.

nicosandller (approved):

PR #57, opened 2026-07-12
Streamline batch: five small fixes (#30 #37 #38 #39 #50)

### LINKED ISSUE #30: Allow user to stretch and shrink rooms
As a user, I want to be able to grab a wall or corner of a room and drag the corner or wall and for the attached walls to get stretched or dragged with it, so that I can more easily expand rooms without having to be very specific in how I draw

### LINKED ISSUE #34: Icons are not top-aligned

### LINKED ISSUE #37: Copy + pasting not working between floors
When I copy an element (furniture, device, etc.), pasting works great on the same floor, but paste does not work as I go to a different floor

### LINKED ISSUE #38: Zoom issue with visual editor
Zooming in the map, causes zoom effect to be applied to entire "visual editor" view, which makes working with the card on smaller devices a bit hard

### LINKED ISSUE #39: Device visibility
When a "device" is not connected to an entity, then the device is not displayed in the map.
Not all my devices have an associated entity in HA, and it would be nice to be able to display these as well.

Might be relevant for another issue, but the device icon is also not always centered, ex. mdi:smoke-detector at 18px.

### LINKED ISSUE #50: Windows shortly behaves like doors
I have a window added to my floorplan.

Whenever I change floors, the window briefly shows the door animation

https://github.com/user-attachments/assets/d09606c3-1456-4178-a22c-c331c702b68f

This is just a visual glitch.

nicosandller:
One PR for the small missing-feature / bug batch. Each fix is independent and verified in the dev harness; unit suite is at 191 after rebasing on main.

> **Note:** this PR originally also fixed #34, but @amadeobriones' #41 landed the same fix first (and better documented — see its measured before/after) — so that part now belongs to #41. This branch has been rebased on top: it keeps #41's version, extends its `gap: 2px` cleanup to the editor, and adds the editor-side counterpart (the canvas `.ilabel` now anchors the same way, so icons sit identically while editing and viewing).

### #37 — copy/paste between floors
Paste itself always worked cross-floor; the bug was focus. Switching floors happens through a `<select>`, which **keeps keyboard focus**, and the editor's typing guard (correctly) swallows shortcuts while a form control is focused — so the Ctrl+V after switching floors died silently. The floor switcher now hands focus back to the canvas. Verified end-to-end with real key events: copy on floor 1 → switch via the select → Ctrl+V pastes onto floor 2.

### #38 — zoom on small devices
The canvas now has real **pinch-to-zoom**: two fingers zoom the plan (centroid-anchored, clamped to the existing 50–300%), and a second finger landing mid-draw cancels the draw instead of leaving a stray wall. Listeners are capture-phase on the canvas wrap so a finger on a badge or handle still counts. Safari's proprietary `gesturestart/change/end` events are canceled — that's the path that zoomed the entire visual editor.

### #39 — devices without an entity
The card filtered out items with no `entity`, so unbound hardware vanished at runtime (it already showed in the editor). They now render as a static badge: icon override or kind default, never highlighted, tap does nothing, no state line. Also from this issue: the icon glyph is now sized to match the badge's pixel parity (shared `itemIconSize`), so e.g. an 18px badge gets a 12px icon (whole-pixel centering) instead of an 11px one sitting on a half-pixel. The default 34px badge keeps its familiar 22px icon. Reconciled with #42: icon precedence is config icon → (no entity → kind default) → registry icon → attribute icon → derived.

### #50 — windows briefly animating like doors on floor switch
Openings were rendered unkeyed, so Lit **reused DOM nodes across floors**: floor B's window inherited floor A's door/window node and the 0.5s leaf/panel transition animated the leftover transform — the flash in the issue's video, and why only *some* windows did it (only those landing on a reused slot). Openings and trackers now render via `repeat()` keyed by id in both card and editor; a floor switch builds fresh nodes, which never transition on first paint.

### #30 — stretch/shrink rooms
Dragging a wall, or one of its corner handles, now carries every **coincident corner** of neighboring walls (tolerance 0.75 units — corner-snapped rooms match exactly), so pulling a wall stretches the room instead of tearing it open. **Alt-drag detaches** the grabbed wall. Group drags are consistent: selected walls translate once, shared corners outside the selection follow. Bonus fix that fell out: corner snapping now excludes the corner being dragged, removing the 26-unit "dead zone" stickiness endpoint drags used to have. Logic lives in `editor-geometry.attachedCorners` with unit tests.

## Verification
- `tsc` clean, **191/191 tests**, vite build 190.9 kB (re-run after the rebase onto #41 + #42).
- Dev harness, all five: cross-floor paste via real keydowns; synthetic 2-finger pinch (zoom 1 → 2 on a 2× spread, gesture canceled, clean teardown); door→window floor switch creates fresh DOM; three items (labeled sensor / bare light / entity-less 18px smoke detector) render with identical badge centers and a 12px icon; corner drag pulls the attached wall, whole-wall drag stretches both neighbors, Alt detaches, undo restores exactly.
- Post-rebase note for reviewers: the auto-merge with #41 silently **duplicated** the `.label` CSS block (no conflict marker); deduplicated in the merge commit, keeping #41's copy.

## Backwards compatibility
No config schema changes. Entity-less items previously hidden at runtime will now appear (that's the point of #39 — but it is a visible change for anyone who left orphaned items in their config). Icon size changes by ±1px at non-default badge sizes.

Closes #30, closes #37, closes #38, closes #39, closes #50.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,8 @@ background):
 
 - **Select** — the default tool. Click an element to select it; Shift/Ctrl-click or drag
   a box to select several at once. Arrow keys nudge the selection (Shift+arrow jumps a
-  full grid cell); **Ctrl/Cmd+C/V/D** copy / paste / duplicate; **Ctrl/Cmd+Z** undoes
+  full grid cell); **Ctrl/Cmd+C/V/D** copy / paste / duplicate — pasting also works
+  across floors (copy, switch floor, paste); **Ctrl/Cmd+Z** undoes
   (**Shift+Z** or **Ctrl+Y** redoes); **Escape** cancels an in-progress draw or clears
   the selection. The **Element** section below the canvas names the selection
   (e.g. *Door · 60 units*) and carries its **duplicate** / **delete** buttons along
@@ -99,6 +100,10 @@ background):
   keeps walls horizontal/vertical and corner-snapped (turn it off to draw freely), and
   the **Snap** segmented control (`On` / `Off` / `Custom`) governs snapping for *all*
   tools — `Custom` lets you snap to a percentage of the grid (e.g. 50% = half a cell).
+  Rooms **stretch** instead of tearing: dragging a wall — or one of its corner
+  handles — carries every wall corner that touches it, so you can widen a room by
+  just pulling its wall. Hold **Alt** while dragging to detach and move only the
+  grabbed wall.
 - **Door / Window** — click to drop; it snaps onto the nearest wall. The context row
   shows a **Length** field for the *next* opening you place, so you can size doors and
   windows before placing them. Assign a sensor after placement (in the **Element**
@@ -117,9 +122,10 @@ background):
 
 Undo/redo buttons sit at the right of the tools row. Zoom controls live on the canvas
 itself (bottom-right): **−** / **+** step, click the percentage to reset, the fit button
-snaps back to 100%, and **Ctrl/Cmd+scroll** zooms from the keyboard/trackpad. The
-**Project** section (canvas size, grid, background) is collapsed by default — click its
-header to expand.
+snaps back to 100%, and **Ctrl/Cmd+scroll** zooms from the keyboard/trackpad. On touch
+screens, **pinch** directly on the canvas — it zooms just the plan (anchored between
+your fingers), not the whole editor. The **Project** section (canvas size, grid,
+background) is collapsed by default — click its header to expand.
 
 ## Elements
 
@@ -149,6 +155,10 @@ By default it shows an icon badge:
   still wins.
 - **Make it yours** — override the **icon** (with autocomplete + live preview), set a
   custom **name**, change the **size**, **rotate** it, or hide the icon entirely.
+- **No entity? Still on the map** — a device with no entity bound renders as a plain
+  badge (its icon override or kind default), so hardware that has no Home Assistant
+  entity — a dumb smoke detector, a wired doorbell — can still be marked on the plan.
+  It never highlights and tapping does nothing.
 
 ### Presence ripples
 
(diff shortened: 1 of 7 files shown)


nicosandller:
@shauneccles can I ask you for a quick review?

amadeobriones (commented):
Reviewed the full diff and both files; ran typecheck + 191 tests (green). Chased the classic failure modes for each fix and none reproduce:

- **#38 pinch-zoom:** centroid math keeps the content point under the fingers; the second-finger handoff cancels an in-progress draw via the capture-phase `stopPropagation`. No stuck state.
- **#37 focus:** confirmed live — with the floor `<select>` focused, switching floors moves `activeElement` back to the canvas, so Ctrl+V lands again. Real fix for a silent swallow.
- **#39 no-entity:** badge renders, never highlighted, tap inert.
- **#30 weld:** both drag paths keep coincident corners together and skip already-selected neighbors.
- The merge resolution (7c27c7d) correctly de-duplicated the `.label` block and kept #41's version.

Optional, none blocking:
- The pointer + `gesture*` listeners are added in `firstUpdated` but not removed in `disconnectedCallback`. Not a real leak (same shadow root, GC's with the instance); a defensive `this._pinchPts.clear()` on teardown would also guard against a lost `pointerup` leaving a stale point that a later tap reads as a phantom second finger.
- `repeat()` keyed by `id` assumes non-empty ids — fine for editor-generated configs, but hand-written YAML omitting `id` yields colliding `undefined` keys; a fallback key would cover imported configs.

Heads-up: now that #60 (rotation) is in `main`, this needs a rebase — it conflicts in `floorplan-card.ts` / `render.test.ts` / README (mechanical rotation overlap, not a logic clash). LGTM once rebased.

nicosandller:
Rebased on main (@amadeobriones — thanks for the thorough review, both suggestions taken):

- **`repeat()` keys** now fall back to the array index (`(o, i) => o.id || i`) in the card *and* editor, so hand-written YAML without ids can't collide on `undefined`.
- **Pinch teardown**: `disconnectedCallback` now clears `_pinchPts`/`_pinch`, which kills the phantom-second-finger scenario you described. One deliberate deviation: the listeners themselves **stay attached**. HA's edit dialog reparents the editor (disconnect → reconnect without a second `firstUpdated`), so removing them on disconnect would permanently disable pinch after a reparent — and since they live on an element inside our own shadow root they die with the instance, as you noted. There's a comment in the code capturing this so nobody "fixes" it later.

Merge with #60 was mostly mechanical (imports, README paragraph, both test suites kept — now **197/197**). The one real composition point: items render **keyed and rotated** (`repeat(..., (it, i) => it.id || i, (it) => this._renderItem(it, c, rot))`). Verified in the harness that the merged behaviors hold together: at `rotation: 90` the entity-less badge remaps to the right percentages inside the swapped viewBox, floor switches still build fresh opening DOM, and pinch state clears on disconnect.

PR #58, opened 2026-07-12
Animate device icons while active (fan spins, playback pulses) — closes #48

### LINKED ISSUE #48: Animated icon when active
Would be a nice touch if when something is active (like a fan) the icon would be animated.

nicosandller:
Implements #48 the way Home Assistant itself animates icons: HA exposes no animation API — its Tile card simply applies CSS keyframes to the icon while the entity is active — so this card now does the same.

## What you get
- **Auto defaults (zero config), matching HA's feel**: a running `fan` **spins**, a playing `media_player` and a cleaning `vacuum` **pulse**. Everything else stays still.
- **Per-device control**: new **Animate icon** dropdown in the element editor (`iconAnimation: auto | none | spin | pulse`) — force a spin/pulse on any entity, or switch it off.
- **Fail-closed, like the active highlight**: an animation is a claim the device is running, so it *only* plays while `entityIsActive` says so — a forced `spin` on an off/`unavailable`/`unknown` entity stays still.
- **Live in the editor too**: the canvas preview animates exactly when the card would, so the dropdown's effect is visible while editing.
- Details: the keyframes ride the inner `ha-icon`, not the badge, so they compose with the user's `angle` rotation; `prefers-reduced-motion: reduce` disables all of it (as HA's frontend does).

## Implementation
- `resolveIconAnimation(item, state)` — pure, shared by card and editor (`src/render.ts`), driven by a small `AUTO_ICON_ANIMATION` domain map.
- `FloorItem.iconAnimation` (`src/types.ts`) — absent field = today's behavior for every domain except the three animated ones; set `none` to restore stillness there.
- Editor form field in `src/editor-forms.ts`; keyframes + reduced-motion guard in both stylesheets.
- Dev harness: `fan.ceiling_fan` (on) and `media_player.living_tv` (playing) mocks.

## Verification
- `tsc` clean, **187/187 tests** (180 on main + 7 new: resolver matrix incl. fail-closed cases, form field/data defaults), build 183.5 kB.
- Harness, computed styles: fan → `fp-icon-spin 2s`, TV → `fp-icon-pulse 1.6s`, forced `spin` on a light animates, `none` on a fan doesn't; identical results in card and editor overlays. Scrubbing the animation clock shows the transform sweep 0°→90°→180° and opacity dip to 0.4; flipping the fan to `off` removes the class live. Editor dropdown round-trips (`none` → `spin` persists to config).

## Backwards compatibility
No schema break. The only visible change without config is that fans / media players / vacuums that are *active* now move; `iconAnimation: none` opts any device out.

Closes #48.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,12 @@ By default it shows an icon badge:
   still wins.
 - **Make it yours** — override the **icon** (with autocomplete + live preview), set a
   custom **name**, change the **size**, **rotate** it, or hide the icon entirely.
+- **Icons that move** — while an entity is active its icon can animate, the way
+  Home Assistant's own Tile card does it: by default a running **fan spins** and a
+  playing **media player** or cleaning **vacuum pulses**. The **Animate icon**
+  dropdown per device switches to `none`, or forces `spin` / `pulse` on any entity
+  (a spinning icon still only plays while the entity is actually active — an
+  unavailable fan never spins). Respects the OS *reduced motion* preference.
 
 ### Presence ripples
 
@@ -397,6 +403,7 @@ distortion. **`imageOpacity`** (0–1, default 1) fades it.
 | `size`        | number                                 | `34`         | Icon badge diameter (px).                              |
 | `angle`       | number                                 | `0`          | Icon rotation (deg).                                   |
 | `display`     | `badge` \| `ripple` \| `iconRipple`    | `badge`      | How the device is drawn.                               |
+| `iconAnimation` | `auto` \| `none` \| `spin` \| `pulse` | `auto`       | Animate the icon while active. `auto`: fan spins; media player / vacuum pulse. |
 | `rippleColor` | string                                 | primary color| Ripple ring color (ripple modes).                     |
 | `rippleSize`  | number                                 | `80`         | Max ripple diameter (px).                              |
 | `showIcon`    | boolean                                | `true`       | Show the icon badge.                                   |
--- a/dev/dev.ts
+++ b/dev/dev.ts
@@ -163,6 +163,19 @@ const hass = {
       state: "off",
       attributes: { friendly_name: "Door Lock", device_class: "lock" },
     },
+    // Icon-animation demo (issue #48): a running fan spins, a playing media
+    // player pulses. Flip their states in the console via setHassBinary /
+    // baseStates to watch the animation start and stop.
+    "fan.ceiling_fan": {
+      entity_id: "fan.ceiling_fan",
+      state: "on",
+      attributes: { friendly_name: "Ceiling Fan" },
+    },
+    "media_player.living_tv": {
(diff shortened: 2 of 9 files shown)


amadeobriones (commented):
Reviewed + ran typecheck + 187 tests (green), and drove the animation in a browser: fan `on`→`anim-spin`, media `playing` / vacuum `cleaning`→`anim-pulse`, all on the inner `ha-icon` so `spin` composes with the badge's `angle` transform instead of fighting it. Fail-closed confirmed — fan `off`, media `paused`, vacuum `docked`, `unavailable` → no animation. Flipping an unrelated entity left an active icon's class unchanged, so no re-render restart churn. Nicely built.

Two small doc fixes (Copilot flagged these too, and they're right): the `IconAnimation` doc in `types.ts` and the editor dropdown label in `editor-forms.ts` both say "fan spins, playback pulses" but omit **vacuum**, which `AUTO_ICON_ANIMATION` does pulse.

One thing worth a note (not necessarily a fix): a `media_player` in state `"on"` (powered, not playing) also pulses — it follows from `entityIsActive` treating `"on"` as active, and it's consistent with the existing `.item.on` highlight, so it's defensible; just flagging so the "pulses while playing" wording isn't misleading.

(Copilot's third point — the `states["undefined"]` lookup — is a non-issue: `FloorItem.entity` is typed required, so the lookup harmlessly misses.)

Heads-up: needs a rebase on `main` now that #60 landed (conflicts in `floorplan-card.ts` / `render.test.ts`, mechanical rotation overlap).

nicosandller:
Rebased on main and review feedback addressed — thanks @amadeobriones for the live drive-through and Copilot for the doc catches:

- **Vacuum is now named everywhere the auto defaults are described**: the `IconAnimation` doc in types.ts, the editor dropdown label ("Auto (fan spins; media & vacuum pulse)"), and the README bullet.
- **The `media_player: "on"` nuance is documented** rather than papered over: the doc and README now say a media player pulses while *playing or simply on*, which is deliberate — it follows `entityIsActive` and matches the existing `.item.on` badge highlight, exactly as flagged.
- **Copilot's `states["undefined"]` point**: agreed it was harmless when `entity` was required — but main has since merged #39 (entity-less items render), so items without an entity genuinely reach `_renderBadge` now. The guard is in (`item.entity ? … : undefined`), matching the editor overlay.

The merge itself was unions all the way down; the one real composition point is the editor badge, which now carries the animation class *and* main's parity icon sizing (`itemIconSize`). Verified the merged surface in the harness with one config combining everything in flight: at `rotation: 90`, a fan spins (`fp-icon-spin`, 22px parity icon) with its `showName` label, the TV pulses, `iconAnimation: none` stays still, and an entity-less badge renders inert. **213/213 tests**, typecheck + build clean.

PR #60, opened 2026-07-12
Rotate the displayed floorplan in 90° steps — closes #33

### LINKED ISSUE #33: Allow user to rotate floorplan
As a user I want to be able to rotate my floor-plan in 90 degree increments, so that I can show it on my dashboard horizontally or vertically

nicosandller:
Implements #33: show a landscape plan on a portrait dashboard (or upside-down, for a wall tablet mounted that way) by rotating the **displayed card** in 90° increments.

## Design
- New top-level **`rotation: 0 | 90 | 180 | 270`** (anything else — 45, negatives, strings — normalizes to 0; negatives/overshoots wrap, so `-90` → 270).
- **Display-only**: no stored coordinate changes, the editor always shows the plan as drawn. Rotating back and forth is lossless by construction.
- **Icons and text stay upright.** Instead of CSS-rotating the whole card (which would tip every badge and label sideways), the SVG layers (walls, openings, furniture, trackers, background image) rotate through a single group transform while the HTML overlay (badges, state labels, text) is remapped point-by-point with the same mapping (`rotatePlanPoint`). A unit test proves the group transform and the point mapping are the same function, and a DOM check confirms it at the pixel level.
- 90°/270° swap the card's aspect ratio (`600 / 1000` instead of `1000 / 600`), so the card takes portrait space in the dashboard grid.
- Wall-gap masking, door swings/slides, tracker animations all ride inside the rotated group unchanged.

## UI (low priority, per the request)
**Rotate display** is the deliberate **last row of the Project section** — implemented as its own one-field form (`projectRotationForm`) so it sorts below the Background and floor-image rows rather than mid-form. Helper text: *"Rotates the live card only — editing stays as drawn."* Selecting 0° removes the key from the YAML.

## Verification
- `tsc` clean, **186/186 tests** (180 on main + 6 new: normalization incl. junk values, size swap, corner/interior point mapping for all four steps, four-quarter-turns-is-identity, transform ≡ point-mapping equivalence, form field/patch behavior).
- Harness: for an L-shaped test room with a door, a lit badge and a text label, all four rotations produce the expected viewBox/aspect/transform and overlay percentages; at 90° and 270° the SVG's own `getScreenCTM()` for the item's plan point matches the badge's DOM center to the decimal. Screenshot at 270° shows the portrait card with the door swung out of the (now-left) wall and the "Kitchen" label upright. Editor canvas stays `0 0 1000 600` while the card rotates; the dropdown round-trips 180 → 0 (key removed).

## Backwards compatibility
No schema break — absent `rotation` renders exactly as today. Nothing else in the config changes meaning.

Closes #33.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,9 @@ Undo/redo buttons sit at the right of the tools row. Zoom controls live on the c
 itself (bottom-right): **−** / **+** step, click the percentage to reset, the fit button
 snaps back to 100%, and **Ctrl/Cmd+scroll** zooms from the keyboard/trackpad. The
 **Project** section (canvas size, grid, background) is collapsed by default — click its
-header to expand.
+header to expand. Its last row, **Rotate display**, turns the *live card* in 90° steps
+for portrait wall tablets — the editor keeps showing the plan as drawn, and icons and
+labels stay upright at any rotation.
 
 ## Elements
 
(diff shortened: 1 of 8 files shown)


nicosandller:
@TruthOf42 can I ask you for a review of this? Since you've got familiarity with the request.

TruthOf42 (approved):
UI/UX rotates and keeps text to bottom of icon

amadeobriones:
Now that this is in — one follow-up I noticed while going through the rotation: `getGridOptions()` and `getCardSize()` still return fixed landscape values, so at 90°/270° the stage becomes portrait (`aspect-ratio: 600 / 1000`) while the card still claims a landscape 12×8 grid slot. In sections view that makes a rotated plan render distorted / mis-sized — it stays internally aligned (both layers stretch together), so it's cosmetic rather than a badge drift, but it's visible. Swapping the reported footprint at 90/270 fixes it.

I've got a small verified patch for it (the exact portrait ratio is really a UX judgment, so it's tunable). Rather than tack scope onto a merged PR, I figured I'd flag it and see how you'd like to handle it — a follow-up PR, an issue, or just leave it for now. No urgency either way.

nicosandller:
Let's create an issue with this @amadeobriones and we can tackle it in another PR!

what do you think?

PR #62, opened 2026-07-13
Items: Show name toggle + adjustable label size — closes #61, #59

### LINKED ISSUE #59: Text, characters
Character height/width not allowed ?! Devicetext height should be equal to angle or editable manualy 

Very good Work so far.

### LINKED ISSUE #61: Show Name instead of State
It would be nice, to have a switch to show the entities name just like we can toggle the state. 
In my case I use a lot of buttons in my HA so i want to build a little panel at the buttom of the floorplan to trigger those. Showing the state is not useful to identify those and adding text labels is a pain when it comes to spacing.

Love the integration :)

nicosandller:
Two small label asks in one PR, since they touch the same line of UI:

## #61 — Show name
New **Show name** toggle per device (`showName`). The label line shows the device's name — the config `name` override, else the entity's friendly name, else the entity id. Built for the reporter's case: a panel of identical button icons where a state label says nothing and text elements are a spacing chore. Combines with **Show state** as `Name · state`; sensors keep their state-shown-by-default behavior, and `showState: false` still silences a sensor while leaving the name.

## #59 — Label size
New **Label size** slider (`labelSize`, default 12px, range 8–40). Shown in the editor only while a label line actually renders (name or state on), and the editor canvas label previews the chosen size live. *Note to reporter:* #59's wording was terse — I read it as "the device label text size isn't editable"; if the ask was actually about label rotation or something else, say so on the issue and we'll follow up.

## Implementation
- `itemBadgeLabel(hass, item)` — one pure helper (`src/render.ts`) builds the label line from both toggles; the card renders whatever it returns, so name/state composition is unit-tested rather than templated.
- `FloorItem.showName` / `FloorItem.labelSize` in types; editor form fields with conditional visibility; README (Devices bullet + item table).

## Verification
- `tsc` clean, **186/186 tests** (180 on main + 6: label composition matrix incl. fallback chain and silenced-sensor case, form field visibility/data defaults).
- Harness: light with `showName` shows "Living Room"; sensor with `showName` shows "Living Area Temperature · 17.9 °C" (HA display precision intact); `name: "Custom Label"` override wins; `labelSize: 20` lands as inline 20px; item with neither toggle renders no label element at all. Editor: "Show name" checkbox commits, the Label size row appears only after a label line exists, and dragging the slider to 18 updates both config and the canvas preview live.

## Backwards compatibility
No schema break; both fields absent = today's rendering exactly.

Closes #61. Closes #59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,10 @@ By default it shows an icon badge:
   still wins.
 - **Make it yours** — override the **icon** (with autocomplete + live preview), set a
   custom **name**, change the **size**, **rotate** it, or hide the icon entirely.
+- **Label line** — **Show state** displays the live value (sensors do this by
+  default); **Show name** adds the device's name — handy for a panel of look-alike
+  buttons where a state would say nothing. Both together read `Name · state`, and
+  **Label size** adjusts the line's font size.
 
 ### Presence ripples
 
@@ -400,7 +404,9 @@ distortion. **`imageOpacity`** (0–1, default 1) fades it.
 | `rippleColor` | string                                 | primary color| Ripple ring color (ripple modes).                     |
 | `rippleSize`  | number                                 | `80`         | Max ripple diameter (px).                              |
 | `showIcon`    | boolean                                | `true`       | Show the icon badge.                                   |
-| `showState`   | boolean                                | sensors only | Show the entity state label.                           |
+| `showState`   | boolean                                | sensors only | Show the entity state in the label line.               |
+| `showName`    | boolean                                | `false`      | Show the device's name in the label line (`Name · state` when combined). |
+| `labelSize`   | number                                 | `12`         | Label line font size (px).                             |
 
 Clicking a `light`, `switch`, `cover`, `fan` or `input_boolean` toggles it; other
 domains open the more-info dialog.
--- a/src/editor-forms.test.ts
+++ b/src/editor-forms.test.ts
@@ -156,6 +156,23 @@ describe("itemForm", () => {
     ).toContain("rippleSize");
   });
 
+  it("offers Show name, and Label size only while a label line renders (#61, #59)", () => {
+    // A light shows no label by default → no size slider.
+    const light = itemForm(item);
+    expect(light.fields.map((x) => x.name)).toContain("showName");
+    expect(light.fields.map((x) => x.name)).not.toContain("labelSize");
+    // Sensors label by default; showName or showState also reveal the slider.
+    const sensor = itemForm({ ...item, entity: "sensor.a", kind: "sensor" } as FloorItem);
+    expect(sensor.fields.map((x) => x.name)).toContain("labelSize");
+    const namedLight = itemForm({ ...item, showName: true } as FloorItem);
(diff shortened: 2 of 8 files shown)


amadeobriones (commented):
The feature is sound — `itemBadgeLabel` composes name/state into a single string (verified live: `showName` off→`"on"`, on→`"Kitchen · on"`), it's a single `<span>` not a second element, defaults preserve existing renders, and it's well unit-tested. Two things worth flagging (a comment, not a block):

**1. Heads-up on resolving the `main` conflict — it's easy to reintroduce the #34/#41 regression.** GitHub already shows this conflicting, so it needs a manual resolution; this is just to make sure it goes the right way. The conflict is the label `<span>`: #41 made it out-of-flow with a `.label.inflow` fallback so a `showIcon:false` item's label still centres on its coordinate; this PR rewrites the same line to add `labelSize`. Taking this PR's side as-is drops `inflow` and label-only items fall 2px off their coordinate — the exact bug #41 fixed. The resolution that keeps **both**:

```ts
${labelText
  ? html`<span class="label ${visual === nothing ? "inflow" : ""}"
      style="font-size:${Math.min(40, Math.max(8, Number(item.labelSize) || DEFAULT_LABEL_SIZE))}px;"
      >${labelText}</span>`
  : nothing}
```

I did the merge in a throwaway worktree to confirm: only `floorplan-card.ts` conflicts, and after this resolution typecheck + the full suite pass.

**2. (hardening) `labelSize` reaches an inline `style` unsanitized** — `font-size:${…}px`. The editor clamps 8–40, but an imported/hand-edited config bypasses the editor. Confirmed live: `labelSize:"20px;color:red"` renders `font-size:20px;color:redpx;` — the payload reaches the `style` attribute (a variant with a sacrificial token last, e.g. `9;position:fixed;inset:0;background:red`, injects cleanly). Clamping/coercing at the sink neutralizes it (with the clamp, both render `font-size:12px;`). Same surface #65 is closing — worth routing through `cssNumber` once that lands, or clamping here. Pre-existing pattern in the codebase, just a natural place to fix it.

LGTM feature-wise; just the resolution care + the sink.

nicosandller:
Rebased on main — @amadeobriones, both points taken, thanks for pre-flighting the merge in a worktree:

1. **Resolution went the way you flagged**: the label span keeps #41's `inflow` fallback *and* gains `labelSize` — verified in the harness that a `showIcon: false` item still renders `class="label inflow"` with `position: static`, centered on its coordinate. The merge also threaded main's #39 guard into `itemBadgeLabel`: no entity → no state line, so an unbound device's label can only be its configured `name` (new unit test).

2. **The sink is clamped**: new `itemLabelSize()` coerces/clamps to the editor's 8–40 range (default 12) at the `style` attribute in both card and editor. Your exact payloads are unit tests now — `"20px;color:red"` and `"9;position:fixed;inset:0;background:red"` both render `font-size:12px;`, confirmed live in the harness too. Treated as interim hardening: once #65's `cssNumber` lands I'm happy to route through it instead.

Suite is at **206/206** after the merge (union with main's rotation/parity/attach tests plus the new guard + clamp cases).

PR #63, opened 2026-07-13
Covers: roll-up opening glyph + safe default tap — closes #45, #47

### LINKED ISSUE #45: Create new icon for covers
As a user, create a separate icon for covers (as opposed to doors) so that when creating garage doors, Im not forced to use a door icon that looks very large and shows a giant door

### LINKED ISSUE #47: Support functionality for Covers
Support for covers would be great:

Image
https://www.home-assistant.io/integrations/cover/

At the moment, clicking on the ICON causes the rolling shutters to close.

nicosandller:
The two cover complaints in one PR, since they're the same story from two angles: covers were drawn like doors (#45) and tapped like lights (#47).

## #45 — a real representation for garage doors and roller shutters
New opening **`motion: "roll"`**. Where a slider's panel travels along the wall, a roll-up cover leaves the floor plane — so the glyph is a **slatted curtain that thins onto its track line** as the cover opens: closed = full-thickness band with slat ticks, 50 % = half-thickness, open = just jambs and a faint track marking the gap. Reads instantly differently from the giant swing leaf the reporter was stuck with, and from the slide panel.

- Partial `cover` positions render proportionally (`scaleY(1 − position)`, 0.5 s transition like the other openings).
- **Device-class inference**: binding a `garage`, `garage_door` or `shutter` cover now sets Roll up automatically ( previously garage inferred *slide*); `blind`/`shade`/`curtain` stay sliders. The Motion dropdown gains **Roll up (garage / shutter)**, and roll hides the hinge/slide/style rows that don't apply.
- The editor previews roll openings partly open (like sliders) so the motion is visible while configuring.

## #47 — tapping a cover icon no longer moves the cover
A bare tap on a cover **item** used to fire `toggle` — the reporter's shutters physically closed from a stray tap. Cover items now default to **more-info**, matching HA's own Tile card. `tap_action: { action: toggle }` remains a one-line opt-in (the #44 action editor exposes it as a dropdown). Deliberately unchanged: clicking the **opening symbol** on the plan still toggles its cover — that's a large, explicit target that exists precisely to operate the cover.

## Verification
- `tsc` clean, **190/190 tests** (180 on main + 10: device-class matrix, motion/default-open/slider-style behavior for roll, glyph structure incl. slat count and absence of slide/swing artifacts at partial amounts, form field visibility + `toPatch`, cover tap default).
- Harness: a `cover.garage` at positions 0/50/100 renders `scaleY(1)` / `scaleY(0.5)` / `scaleY(0)` with 9 slat ticks on a 120-unit opening; tapping the cover item fires `hass-more-info` for `cover.garage` with **zero** service calls; editor Motion dropdown offers Roll up, selecting it patches `motion: "roll"` and clears `sliderStyle`, and the canvas previews the half-rolled curtain (`transform-box: fill-box` confirmed so it collapses onto the track, not the SVG origin).

## Backwards compatibility
- Existing configs render unchanged (`roll` is new; no stored value is reinterpreted). Newly *bound* garage/shutter entities infer roll instead of slide — existing openings keep whatever motion they have.
- **Behavior change to note in the release**: cover item taps now open more-info instead of toggling. This is the point of #47, but anyone relying on tap-to-toggle covers needs the one-line `tap_action` opt-in (README updated).

Closes #45. Closes #47.

## Repro in the dev harness (no Home Assistant needed)

The branch now ships mock roll-up covers so anyone can exercise the feature locally:

```
npm install && npx vite
# open http://localhost:5173/dev/
```

**Interactive path (zero code edits):**
1. Draw a wall, drop a **Door** on it.
2. In the **Element** section, set **Entity** to `cover.garage_door` (or `cover.bedroom_shutter`).
3. The editor infers **Motion: Roll up** from the device_class on its own — watch the Motion dropdown flip.
4. An **opening emulator** panel appears below with a 0–100 position slider for the cover; drag it and the slatted curtain thins/thickens on the live card (`scaleY(1)` closed → `scaleY(0.4)` at 60 → `scaleY(0)` open, with the 0.5 s transition).

**Demo-floor path:** flip `START_WITH_DEMO` to `true` in `dev/dev.ts` — the sample room now includes a garage door pre-bound to `cover.garage_door` on the east wall, so the slider is live immediately on load.

The tap-safety half (#47) is reproducible the same way: add a **device** bound to `cover.garage_door` and tap its badge — the mock `hass` logs show no `callService`, and a `hass-more-info` event fires instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,9 @@ A **device** binds a Home Assistant entity to a spot on the plan. Add one with
 **+ device**, then pick the entity in the **Element** section below the canvas.
 By default it shows an icon badge:
 
-- **Tap to act** — lights, switches, covers, fans and `input_boolean`s toggle on tap;
+- **Tap to act** — lights, switches, fans and `input_boolean`s toggle on tap;
+  covers open the more-info dialog instead (an accidental tap must not move a
+  shutter — set **Tap action** to *Toggle* to opt back in);
   any other entity opens its more-info dialog.
 - **Live look** — the badge highlights when the entity is "on". Turn on **Show state**
   to display the current value next to it, formatted exactly as Home Assistant would —
@@ -177,8 +179,9 @@ Select the opening and bind an **Entity** in the **Element** section below the c
 contact `binary_sensor` or a `cover` (Home Assistant's domain for anything that opens: doors,
 gates, garages, blinds, shades, shutters, curtains…) — to make the opening track its real
 state. When you bind an entity the card reads its HA **`device_class`** and sets a sensible
-`type`/`motion` for you (a `window` cover → a window; a `garage` roller → a sliding door);
-adjust either afterwards. Once bound, the opening tracks state:
+`type`/`motion` for you (a `window` cover → a window; a `blind` → a slider; a `garage`
+or `shutter` roller → a **roll-up**); adjust either afterwards. Once bound, the opening
+tracks state:
 
 - **Open / closed** — the opening is drawn open when the entity is `on` / `open`, closed
   otherwise. A door's leaf swings around its hinge; a window's two leaves swing outward
@@ -372,7 +375,7 @@ distortion. **`imageOpacity`** (0–1, default 1) fades it.
 | ------------- | --------------------------- | ------------------------------------------------------ |
 | `id`          | string                      | Unique id.                                             |
 | `type`        | `door` \| `window`          | The kind of opening.                                   |
-| `motion`      | `swing` \| `slide`          | How it moves. `swing` (default) hinged door / casement window; `slide` sliding panels. |
+| `motion`      | `swing` \| `slide` \| `roll` | How it moves. `swing` (default) hinged door / casement window; `slide` sliding panels; `roll` roll-up cover (garage / roller shutter) drawn as a slatted curtain that thins onto its track as it opens. |
 | `x`, `y`      | number                      | Center position.                                       |
 | `length`      | number                      | Length along the wall.                                 |
 | `angle`       | number                      | Rotation in degrees.                                   |
--- a/dev/dev.ts
+++ b/dev/dev.ts
@@ -163,6 +163,30 @@ const hass = {
       state: "off",
       attributes: { friendly_name: "Door Lock", device_class: "lock" },
     },
+    // Roll-up covers (issue #45): binding either to an opening infers
+    // motion "roll" from the device_class (garage → door, shutter → window)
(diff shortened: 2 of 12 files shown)


nicosandller:
@TruthOf42 also requesting a review of this due to familiarity if time allows!

Thanks!

TruthOf42:
@nicosandller is it possible to add test data for showing roll-up cover in the branch, or include it when in dev mode, or other repro steps?

amadeobriones (commented):
Reviewed this closely — checked out the branch, ran typecheck + the full suite (190 pass), and drove the actual render in a browser since the committed tests assert on serialized markup.

**Roll glyph is solid.** Every cover state — closed→`scaleY(1)`, 50%→`0.5`, 100%→`0`, position-less falling back to binary open/closed, `opening` using position, `unavailable`→closed — no NaN anywhere. The band rect is centred on the track line, so the `scaleY` collapse lands right whether or not `transform-box: fill-box` resolves. Robust.

**The badge-tap safety fix works** — tapped a `device_class: garage` cover item, it fired `hass-more-info` with zero service calls. 👍

One scoping note on "safe default tap": it fixes the item-badge tap, but the opening-symbol tap (`_onOpeningClick`) still fires an unconfirmed `cover.toggle` on any open/close-capable cover, garage doors included — no `confirmation`, no `device_class` distinction (confirmed live). That's pre-existing, not introduced here, so this PR doesn't regress anything — but since it's titled "safe default tap," it might be the natural place to either extend confirmation/device-class awareness to `_onOpeningClick` or scope the wording to the badge. Your call.

On "closes #47": the up/down/stop ask is met indirectly via more-info rather than in-card controls — worth noting on the issue. Nice-to-have: a render test for a position-less and an `unavailable` roll cover (only slide/swing have amount-fallback coverage today). LGTM overall.

nicosandller:
> @nicosandller is it possible to add test data for showing roll-up cover in the branch, or include it when in dev mode, or other repro steps?

added a little something on the dev harness. 

Give it a try!

nicosandller:
Rebased on main. One real conflict — this branch's roll-up cover mocks and #58's animation mocks landed on the same lines of the dev harness states; resolved as a union (both demo sets kept). Verified the merged surface in the harness: the garage curtain renders `scaleY(0.5)` at position 50, tapping the cover item fires `hass-more-info` with zero service calls, a #62 `showName` label and a #58 spinning fan render beside it, and cover icons stay still (they're not in the auto-animation map). **223/223 tests**, typecheck + build clean.

Issue #64, opened 2026-07-13
Harden config colour/size values against style-attribute injection


amadeobriones:
> **🤖 Disclosure** — written with an AI agent (Claude Code) under my direction. I run this card on a live HA instance daily and verified the behaviour below against current `main`.

A few user-supplied config fields are interpolated straight into `style="…"` attributes, and Lit doesn't escape `;` or `}` inside a style-attribute binding — so a value can break out of its declaration into arbitrary CSS. Flagging it because the card is designed to be shared/imported, which is the threat model.

**Where (current `main`):**
- the stage `background` — `style="…; background:${c.background …}"`
- an opening's `activeColor` — flows into `style="stroke:${tone}"` when the opening is sensor-active
- label / text `font-size` — the numeric `${t.size}px` (and #62 adds a `labelSize` field in the same shape)

**Impact** (the shape, not a step-by-step): a shared/imported config could set e.g. `background: "red;position:fixed;inset:0;z-index:99999"` → a full-viewport overlay over the HA UI, or a colour containing `url(…)` → a remote fetch (an IP-leaking beacon, which HA cards otherwise avoid).

**Not affected:** the SVG `fill=`/`stroke=` *presentation attributes* (walls, furniture strokes) — those aren't `style=`, so there's no declaration to break out of.

**Fix:** validate these before they reach the attribute — an allowlist for colours (accept hex / named keyword / `rgb()`/`hsl()` / `var(--…)`, drop anything else and fall back to the default) and a numeric coercion for the size fields. I ran into this and hardened it on my side that way (fuzzed it hard, no bypass) — happy to open a small PR with the sanitiser plus the wiring at each sink if useful. Review-as-safety-net; flagging so it doesn't go unaddressed.

PR #65, opened 2026-07-13
Harden config colour/size values against style-attribute injection

amadeobriones:
> **🤖 Disclosure** — written with an AI agent (Claude Code) under my direction. I run this card on a live HA instance daily. Addresses #64.

Config is shareable/importable, so its colour and size fields are effectively attacker-controlled. Lit doesn't escape `;` or `}` inside a style-attribute expression, so a shared/imported config can break a value out of its declaration — e.g. `background: "red;position:fixed;inset:0;z-index:99999"` paints a full-viewport overlay over the HA UI, and a colour containing `url(…)` becomes an IP-leaking remote fetch.

**What this does**
- Adds `src/css-safe.ts`: an **allowlist** `cssColor`/`cssColorOr` (hex, named + CSS-wide keywords, `rgb/hsl/hwb/lab/lch/oklab/oklch/color(...)`, and `var(--x[, fallback])`) plus a `cssNumber` coercion for numeric fields. Allowlist over denylist on purpose — a denylist invites the one vector you forgot.
- Wires it at **every `style="…"` sink** in the card, the editor preview, and the shared render helpers: stage `background`, text `color`/`size`, item badge `size`/`angle`, opening `accent` (the `fill`/`stroke` in `style=`), and ripple `color`/`size`.
- Leaves SVG `fill=`/`stroke=`/`transform=` **presentation attributes** alone — there's no CSS declaration there to break out of, so they're not this bug class.

**Honest scoping**
- The colour allowlist is the part I've run hard; the `size`/`angle` numeric coercion is a simple finite-number guard (new, not previously battle-tested, but small and covered by tests).
- Top-level `width`/`height`/`grid` already go through your `setConfig` numeric check, so those are left as-is.
- #62 adds a `labelSize` field of the same shape; it's not on `main` so it's out of scope here — it just needs the same `cssNumber` wrap wherever it lands.

**Backward-compatible:** every legitimate value renders unchanged — I tested the allowlist against the colours real configs use (HA theme vars with/without fallback, hex, named, rgb/rgba in comma and modern space syntax, hsl/hsla, oklch/lab/lch/hwb/color). Only invalid/malicious values fall back to the existing defaults. `tsc` clean; full suite green (adds sanitiser unit + adversarial/fuzz cases).

Marked as a draft — happy to adjust scope, naming, or split it if you'd prefer smaller.

### DIFF
--- a/src/css-safe.adversarial.test.ts
+++ b/src/css-safe.adversarial.test.ts
@@ -1,0 +1,105 @@
+import { describe, it, expect } from "vitest";
+import { cssColor } from "./css-safe";
+
+describe("cssColor — adversarial: exotic breakout attempts all rejected", () => {
+  const ATTACKS = [
+    // newline / CR / tab tricks (the classic ^...$ before-\n regex hole)
+    "red\n;position:fixed",
+    "red\r\n}body{x:1",
+    "red\t;color:blue",
+    "rgb(0,0,0)\n;position:fixed",
+    "var(--x)\n;evil",
+    "red ;evil", // line separator
+    "red;evil", // NEL
+    "red;evil", // vertical tab
+    "red\f;evil", // form feed
+    // CSS comments
+    "red/**/;position:fixed",
+    "rgb(0,0,0)/*x*/",
+    "/*x*/red",
+    // CSS escapes / encoded
+    "\\72 ed;evil",
+    "\\0000red",
+    "red\\3b evil",
+    // url / fetch vectors in every wrapper
+    "url(https://evil/x)",
+    "URL(//evil)",
+    "image-set(//evil)",
+    "var(--x, url(//evil))",
+    "rgb(0,0,0);background:url(//evil)",
+    "expression(alert(1))",
+    "EXPRESSION(1)",
+    // breakout punctuation
+    "red;",
+    "red}",
+    "red{x:1}",
+    "red !important;x:1",
+    "rgb(0,0,0) ;evil",
+    "#fff;--x:url(//evil)",
+    // nested / doubled parens (where url could hide)
+    "rgb(rgb(0,0,0))",
+    "hsl(calc(1) 0 0)",
+    "var(--x, rgb(0,0,0); z:1)",
+    // whitespace-obfuscated
+    "red ; position : fixed",
+    // quote / angle-bracket escape attempts
+    'red"x',
+    "red'x",
+    "red>x",
+    "red<x",
+  ];
+  for (const a of ATTACKS) {
+    it(`rejects ${JSON.stringify(a)}`, () => {
+      expect(cssColor(a)).toBeUndefined();
+    });
+  }
+
+  // Surrounding whitespace (incl. newlines/tabs) is trimmed, and the RESULT is
+  // still validated — so a whitespace-wrapped legit colour is accepted and its
+  // output carries no control char. This is safe, not a bypass.
+  it("trims surrounding whitespace/newlines to a clean value", () => {
+    expect(cssColor("red\n")).toBe("red");
+    expect(cssColor("\nred")).toBe("red");
+    expect(cssColor("  #03a9f4\t")).toBe("#03a9f4");
+    // but an INTERNAL newline followed by a payload is still rejected
+    expect(cssColor("red\n;evil")).toBeUndefined();
+  });
+});
+
+describe("cssColor — INVARIANT: nothing accepted can break out of a style declaration", () => {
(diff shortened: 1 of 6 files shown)


amadeobriones:
Good catch from the Codex pass — the follow-up commit addresses it, and it turned out to point at something broader.

The nested `var()` fallback it flagged (`var(--a, var(--b, #fff))`) was indeed being rejected. Digging in, the same limitation also rejected `rgb(var(--rgb-primary-color))` / `rgba(var(--rgb-*), …)` — the standard way HA has expressed theme colours since 2022.12 (stored as bare `r, g, b` triplets). So the first pass would have silently reverted a lot of real theme values to their defaults.

Rather than special-case those, I swapped the fixed colour-regexes for a structural check: allowed characters only (no `;` `{` `}` `"` `'` `:` `@` `\` `!`), balanced parens, and every function call on an allowlist of inert ones (colour / gradient / `var` / `env` / maths). Arbitrary nesting like `rgb(var(--x))` and `var(--a, var(--b, #fff))` now passes, while `url()` / `image-set()` / `expression()` / `attr()` / `element()` / `paint()` can never appear, nested or not.

I checked it against theme-colour forms from across HA versions — named, hex, `rgb()`/`hsl()` literals, traditional `var(--…)`, the older `--paper-*` / `--mdc-theme-*` variables, the 2022.12+ `rgb(var(--rgb-*))` idiom, fallback chains, and modern `color-mix()` / `light-dark()` — plus an adversarial + fuzz battery. Still O(n), no ReDoS.

One trade-off worth flagging: it's slightly more permissive than a strict grammar — an invalid value like `#xyz` is accepted and simply ignored by CSS rather than pre-rejected. Safe (no breakout), and it's what lets the full HA colour syntax through.

amadeobriones:
Following up on my own PR with a gap I found while re-auditing this.

The colour/size sanitiser is solid, but `width`/`height` still reach the stage's `aspect-ratio` **raw** — `aspect-ratio: ${c.width} / ${c.height}` at `editor.ts` and `floorplan-card.ts`. The card's `setConfig` guards those as numbers, but the editor's `setConfig` doesn't, so a config imported/opened in the editor can break out. Reproduced in the dev harness: setting `width` to `"1 / 600; position: fixed; inset: 0; z-index: 99999; background: rgb(255,0,0)"` makes the editor `.stage` compute `position: fixed` — the injected CSS takes effect (a full-viewport overlay; swap for `url()` via the height field and it's a remote fetch).

Fix is ready and verified: coerce both sinks through `cssNumber(c.width, DEFAULT_WIDTH)` / `cssNumber(c.height, DEFAULT_HEIGHT)`, plus `cssNumber` should treat a blank string as unset (`Number("")` is `0`). With it applied, the same payload renders `aspect-ratio: 1000 / 600` and `position` stays `relative`; adds regression tests; full suite green (309). Happy to push it to this branch — flagging here first so it's visible.

nicosandller (approved):
**Approved — and we agree with the scope as drawn**: style-attribute sinks only, SVG presentation attributes deliberately excluded (no declaration to break out of there), numeric fields coerced rather than pattern-matched, allowlist over denylist. That's the right boundary for this bug class, and the test suite — legit HA colours across every theme era, the hand-written attack corpus, and the 500-case fuzz invariant — is the most thorough in the repo.

I've rebased the branch on main (pushed to your fork — hope that's welcome; it was the same mechanical churn as the other rebases this week):

- **One conflict**, the badge `size` line vs #58's animation resolver — resolved as a union, so the sanitized size and the animation class both stay.
- **#62's `labelSize`** now routes through `cssNumber` exactly as your PR body anticipated: `itemLabelSize` keeps its 8–40 range clamp but delegates coercion to your shared guard, so every numeric style sink uses one primitive.
- Audited the merged tree for style sinks that arrived after your branch (#57/#58/#60/#62) — all covered; no unsanitized interpolation remains.

Verified end-to-end: `tsc` clean, **341/341 tests**, build clean, and a live harness pass — the demo attack (`red;position:fixed;inset:0;z-index:99999` as `background`) falls back to the theme default, a `url(…)`-smuggling text colour is neutralized, hostile `size`/`labelSize` strings coerce to defaults, while `var(--primary-color, #03a9f4)` and the `rgba(var(--rgb-primary-color), 0.5)` triplet idiom render byte-identical and the merged features (icon animation, labels, rotation) still work.

Two style sinks in my still-open branches (#63's roll curtain `scaleY`, #69's editor preview) will adopt these helpers when they rebase after this lands. Ready to merge.

PR #92, opened 2026-07-30
Color elements by entity state (#68, #79, #82)

### LINKED ISSUE #68: Dynamic color for sensor state values
Hi,

First of all, thank you for creating Easy Floorplan. It's a great and very intuitive floor plan card for Home Assistant.

I would like to suggest a feature that would greatly improve the visualization of sensor values.

Use case

I display room temperatures using sensor items:

items:
  - entity: sensor.living_room_temperature
    kind: sensor
    showState: true
    showIcon: false

It would be very useful if the displayed state text could change its color depending on the sensor value.

For example:

Below 24 °C → white
24–26 °C → orange
Above 26 °C → red

This would make it possible to identify critical values directly from the floor plan without having to read every number.

Possible implementation

One possible approach could be a new property like:

stateColor:
  - above: 26
    color: red
  - above: 24
    color: orange
  - default: white

or alternatively a JavaScript template similar to:

stateColorTemplate: |
  if (parseFloat(state) >= 26) return "#ff0000";
  if (parseFloat(state) >= 24) return "#ff9900";
  return "#ffffff";

The same concept could also be useful for humidity, CO₂, battery level, power consumption and many other sensor types.

Thank you for considering this feature!

### LINKED ISSUE #79: Feature Request: Custom active state colors per device type or entity
Feature Request:
Currently, when devices/entities are active (ON), all active badges/icons default to yellow. When many different device types are present (lights, covers, switches, thermostats), it is hard to visually distinguish between them at a glance.

It would be great to have an option in the element/card configuration to customize the active state color:
- Either **globally per domain / device type** (e.g., Lights = Yellow, Covers/Rolläden = Purple/Lila, Thermostats = Orange, Switches = Blue)
- Or **individually per device element** in the visual editor.

This make the great floorplan much more intuitive to read.

Image

### LINKED ISSUE #82: entity for plants
Hi, 
I have soil sensors for my plants. It would be cool if it would be possible to assign entities to plants and using the solution from #79 to color them (e.g.<50%=red, <65%=yellow, <80% green).

nicosandller:
Closes #68, closes #79, closes #82.

These three issues are the same feature seen from three angles — *an element's color should follow its entity* — so they share one mechanism instead of getting three.

## What each issue needed

**#68 (threshold colors) was already half-built.** `StateColorRule`, `resolveStateColor` and `FloorItem.stateColor` all existed and were wired into the card; what was missing was the editor UI, exactly as the issue thread says. So #68 cost a UI, not an engine — and that UI is what the other two reuse.

**#79 (per-device active colors)** was the smallest. `Opening.activeColor` already existed; `FloorItem` just never got the same field. The hard-coded yellow becomes a `--fp-active` custom property with today's value as the fallback, so a device that sets nothing renders byte-for-byte as before.

**#82 (plants)** needed `entity` on `Furniture` — the one placeable element that couldn't bind one. `renderFurniture` funnels every shape through a single `color` local, so binding it was one line rather than one edit per furniture type. Plants and fish tanks are furniture, per the issue discussion.

## The shared piece

`StateColorRule` gains an optional `state`, so one rule type covers numeric thresholds (#68, #82) and word-valued entities (#79):

```yaml
stateColor:
  - state: open      # exact match wins
    color: red
  - above: 26        # then highest matching threshold
    color: orange
  - color: white     # then the default rule
```

Purely additive — existing `above`-only configs keep their behaviour, and the pre-existing tests prove it unchanged.

The editor gets one **Color by state** block, used by both devices and furniture. It renders as plain rows rather than `ha-form` fields: the list is repeatable and `ha-form` has no selector for that (`object` is a raw YAML box). Given #91, worth being explicit — colors are the one part of this editor that was *always* hand-rolled, so the block matches its neighbours by construction. The eight copy-pasted color rows collapse into a single `_renderColorRow`, which is what pays for the ninth.

## Security

Every new color path runs through the `cssColor` allowlist (#64) before reaching a style attribute or an SVG `stroke` — `resolveStateColor` deliberately returns config strings unfiltered, so each caller is gated and the adversarial suite now covers both new ones.

## Verified

460 tests pass (+21), typecheck and build clean. In the dev harness, with a soil-sensor stub and an entity-bound plant added:

- the plant recolours live as the sensor moves — 90 green, 70 amber, 40 red — and the threshold is strict (81 green, 80 amber). This also proves the `collectWatchedEntities` wiring; without it the card would never re-render for the sensor at all, which was the easiest thing in this PR to miss.
- two active devices with different `activeColor`s render `#ffb300` and `#8e24aa` while an untouched active light stays the original `#fdd835` — no regression for existing configs.
- editing a rule updates the live card immediately; a hostile color (`red;position:fixed`) is rejected and the drawing falls back to gray rather than injecting.
- no row overflows at a 366px panel, the realistic docked-inspector width.

The harness has no `ha-form`, so as always the in-HA rendering of the *form* fields can't be exercised locally — but the rules block adds no new `ha-form` fields, so that gap is narrow here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,12 @@ By default it shows an icon badge:
   of the state (a climate's `current_temperature` rather than "heat"), and
   **2nd attribute** for a second reading from the same device — so one climate
   entity renders `21.5 °C · 45%`. Formatted by HA's own attribute formatter.
-- **Threshold colors** — `stateColor` (YAML) colors the label by value:
+- **Active color** — every active device is the theme's yellow by default, which
+  makes a wall of lights, covers and switches hard to read. Set **Active color**
+  per device to tell them apart at a glance — lights yellow, covers purple,
+  thermostats orange. Ripples follow it unless you set a ripple color too.
+- **Color by state** — the **Color by state** rules color the label by what the
+  entity reads. Add rules in the editor, or in YAML:
 
   ```yaml
   stateColor:
@@ -203,7 +208,18 @@ By default it shows an icon badge:
   ```
 
   The highest matching `above` wins; non-numeric values use the default rule.
-  Colors go through the same injection allowlist as every other config color.
+  A rule can also match an exact state instead of a threshold, for entities
+  whose value is a word rather than a number:
+
+  ```yaml
+  stateColor:
+    - state: open
+      color: red
+    - color: green
+  ```
+
+  An exact `state` match beats a threshold. Colors go through the same
+  injection allowlist as every other config color.
 - **No entity? Still on the map** — a device with no entity bound renders as a plain
   badge (its icon override or kind default), so hardware that has no Home Assistant
   entity — a dumb smoke detector, a wired doorbell — can still be marked on the plan.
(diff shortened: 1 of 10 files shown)


DiRe-Code:
Thank you. I Like it.

IMG_0529

PR #107, opened 2026-08-02
Areas: conditional fill and outline coloring from a bound entity (issue #6)

### LINKED ISSUE #6: Area conditional coloring
It would be nice to be able to have different colors for each room.

Also to have that color conditional to temperature / lights and / or time of the day.

superflyer11:
Closes #6.

Areas could only take a static `color`/`opacity`, so a room couldn't react to anything. `Furniture` already solved exactly this for #82 with `entity` + `stateColor` + `activeColor` — this applies the same pattern to `Area`, so a room can colour its **fill**, its **outline**, or both from a bound entity.

### What it adds

| Field | |
|---|---|
| `entity` | Bind any entity to the area |
| `stateColor` | Threshold/state rules, same shape as a device's |
| `activeColor` | Color while the entity is active, when no rule matches |
| `activeOpacity` | Fill opacity while live, falling back to `opacity` |
| `borderColor` / `borderWidth` | Static outline — none is drawn by default |
| `highlight` | Where a live color paints: `fill` (default), `border`, `both` |

Two additions beyond furniture's set, both earning their place:

- **`activeOpacity`** — otherwise you must choose between a resting fill strong enough that you notice it change, and one subtle enough not to dominate the plan. A room can sit at `0.12` and lift to `0.35` when it goes active.
- **`highlight`** — filling a whole room is heavy on a busy plan. `highlight: border` lets a room outline itself while occupied without tinting everything inside it. `activeOpacity` stays a fill concern and deliberately does not apply in that mode.

### Implementation

`areaColor()` mirrors `furnitureColor()`: `stateColor` rules win, then `activeColor` while the entity is active, else `undefined` so the static `color` still applies. It returns a value already through the style-injection allowlist (#64), since it flows straight into `fill`/`stroke` attributes.

`renderArea(a, liveColor?)` takes the resolved colour as an optional second argument, so the editor's existing `renderArea(a)` call is untouched. **An area with no entity, no border and no highlight renders exactly as before** — `stroke=none`, `stroke-width=0`.

The editor gets an entity picker on the area form, placed last, matching furniture's reasoning that most areas are just outlines. Colours stay YAML-only, as furniture's do.

### Example

```yaml
areas:
  # Fill lights up green while occupied, and lifts to a stronger fill.
  - id: kitchen
    entity: binary_sensor.kitchen_occupancy
    activeColor: "#4caf50"
    opacity: 0.12
    activeOpacity: 0.35

  # Outline-only: draws a green border while occupied, fill never changes.
  - id: hall
    entity: binary_sensor.hall_occupancy
    activeColor: "#4caf50"
    highlight: border
    borderWidth: 4

  # Threshold a numeric sensor, so the room reddens as air quality drops.
  - id: study
    entity: sensor.study_co2
    stateColor:
      - { above: 1200, color: "#e1243b" }
      - { above: 800, color: "#ff9300" }
      - { color: "#58d32f" }
```

### Tests

16 new tests: rule precedence, the inactive and unbound cases, the css-safe gate on both `activeColor` and `borderColor`, the `activeOpacity` switch, the no-outline default, the border-width fallback, each `highlight` mode, and a live colour overriding a static `borderColor`.

Full suite green at **499 passing**; `tsc --noEmit` and `vite build` both clean.

Happy to rename anything, or to split the outline work out if you'd rather take the fill on its own first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ automatically to the card and screen size.
 - **Areas** — trace a room's outline point-by-point (points snap to wall corners and to
   neighboring areas' corners) to get a colored, named room polygon. Bind an entity to a
   room and its fill goes live — green while a presence sensor is occupied, red when a CO2
-  sensor crosses a threshold. Naming a room after
+  sensor crosses a threshold. Or bind a **light** and the room glows with that light's own
+  color and brightness, for a bird's-eye view of the tones set across the house. Naming a room after
   one of your Home Assistant areas (the name field autocompletes against them) links the
   two, which scopes the entity picker — for any device dropped inside it — to that HA
   area's entities, and can bulk-add every device in the area.
@@ -681,7 +682,7 @@ trackers:
 
 ### Area
 
-`{ id, points, name?, showName?, color?, opacity?, haArea?, filterEntities?, entity?, stateColor?, activeColor?, activeOpacity?, borderColor?, borderWidth?, highlight? }`
+`{ id, points, name?, showName?, color?, opacity?, haArea?, filterEntities?, entity?, lightEntity?, stateColor?, activeColor?, activeOpacity?, borderColor?, borderWidth?, highlight? }`
 
 - `points` — an array of `{ x, y }` vertices (canvas units), in drawing order; the shape
   is implicitly closed from the last point back to the first.
@@ -696,9 +697,12 @@ trackers:
   a linked `haArea`.
 - `entity` — optional entity that makes the room itself live, in the same shape furniture
   uses. Drives `stateColor` and `activeColor`; an unbound area stays a static polygon.
+- `lightEntity` — a light in this room, whose own color and brightness paint it. See
+  **A room lit by its light** below.
 - `stateColor` — threshold/state rules for the fill (same shape as a device's
-  `stateColor`). Evaluated against `entity`'s state; takes precedence over `activeColor`
-  and `color`.
+  `stateColor`). Evaluated against `entity`'s state. A rule naming an `above` threshold
+  or an exact `state` outranks `lightEntity`; a catch-all rule (neither) ranks *below*
+  it — see the precedence list below.
 - `activeColor` — fill color while `entity` is active, used when no `stateColor` rule
   matches.
 - `activeOpacity` — fill opacity while `entity` resolves a color. Lets a room lift out of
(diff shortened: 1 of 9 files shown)


nicosandller:
Thanks @superflyer11 — this is careful work, and mirroring `furnitureColor` rather than inventing a new shape was the right instinct. I've pushed three changes onto the branch rather than leaving them as review comments, since two of them are intertwined. Happy to revert any of it if you disagree.

### 1. A bug that made the feature inert

`collectWatchedEntities` never collected `areas[].entity`. The card's `shouldUpdate` drops any pure-`hass` tick whose changed entity isn't in that set, so a bound area was painted once and then **frozen** — it only came alive if something *else* on the plan happened to watch the same entity.

I isolated it in the harness by holding everything constant except that:

| Plan | Watched set | Fill when the sensor goes `on` |
| --- | --- | --- |
| Area bound to occupancy, nothing else | `[]` | **unchanged** |
| Same, plus a device on the same entity | `[binary_sensor.kitchen_occupancy]` | ✅ repaints |

This is the same bug the furniture loop immediately above it was added to fix in #82 — its comment even describes the symptom ("the plant stays its first-painted color forever"). Worth noting *why* the 16 tests passed anyway: they all exercise `areaColor`/`renderArea` in isolation and never cross the card's update gate, which is exactly where this class of bug lives. There's now a `collectWatchedEntities` test.

### 2. `lightEntity` — a room lit by its light

From the issue thread: for a **light**, you don't really want conditional rules, you want the light's own colour. So a room can now bind a light directly and take its `rgb_color` and `brightness`.

The important detail is that reading `rgb_color` alone would do nothing in most rooms. Sampling a real install:

| `supported_color_modes` | Count (of 8 sampled) | Can report colour? |
| --- | --- | --- |
| `color_temp` / `xy` | 3 | ✅ |
| `brightness` | 3 | ❌ |
| `onoff` | 2 | ❌ |

So it degrades in rungs: colour-capable → paints its colour; brightness-only → keeps the room's own colour and varies opacity; on/off → fully lit; off/`unavailable` → paints nothing and the room rests (failing closed, like the other state readers here). HA derives `rgb_color` even for `color_temp`-only bulbs, so warm white still reads as amber.

Brightness maps into a **0.15–0.5** opacity band rather than 0–1 — a room dimmed to 10% is invisible, and "I can't see the room" reads worse than "the room is dim".

### 3. Precedence, and a trap in your example configs

The chain is: **specific rule → light → `activeColor` → catch-all rule → static `color`.**

The subtlety: a `stateColor` catch-all (neither `above` nor `state`) matches *everything*. Had it kept top priority, it would have silently masked the light on every config that has one — and every example in your README ends with a catch-all. So a rule that actually **tested** the value outranks the light (a smoke alarm shows through a lit room), while a catch-all ranks below it, since it really means "resting colour". `matchStateColor` now reports which tier matched; `resolveStateColor` keeps its old signature and behaviour.

### Smaller things

- `renderArea` now takes an `AreaPaint` (`color` and/or `opacity`) instead of a bare colour, so a brightness-only light can light a room in the room's *own* colour. An opacity-only paint never draws an outline, having no colour for one.
- A rejected `borderColor` no longer gets a nonzero `stroke-width` (it landed as `stroke="none"` with a width).
- An unsafe rule colour now falls *through* to the light instead of swallowing the rung.

### Verification

Harness-tested against the attribute shapes of real lights — colour / dimmed / brightness-only / on-off / off / unavailable each paint as intended — plus a full lifecycle driven by **pure `hass` ticks**, which is what change 1 buys:

| Step | Fill |
| --- | --- |
| light off, no smoke | `#3366cc @ 0.12` |
| light **on** | `rgb(255, 170, 80) @ 0.5` |
| + smoke detected | `#e1243b` |
| smoke clears, still lit | `rgb(255, 170, 80) @ 0.5` |
| light off again | `#3366cc @ 0.12` |

**518 tests** pass, `tsc --noEmit` and `vite build` clean.

Your `activeOpacity` and `highlight` additions both survive untouched — the reasoning you gave for each holds, and `highlight: border` in particular is the right call for a busy plan.

nicosandller:
**Correction to my previous comment** — the `lightEntity` field I described there is gone.  I revisited the design and landed somewhere better. The rest of that comment still stands (the watched-entities fix, and your `activeOpacity` / `highlight` work).

### What changed

Binding a light to a whole **area** tints the entire room one colour. That can't express an open-plan space, or one lamp warm and another cool at opposite ends of a living room. Light belongs where the lamp is.

So the light colour moved onto the **device**: `glow: true` on a light, and it casts a pool from its own position.

```yaml
- id: lamp
  entity: light.living_standing_lamp
  kind: light
  x: 400
  y: 300
  glow: true
  glowRadius: 200
```

This turned out to need far less machinery than the area version. A device already has an entity and `x`/`y`, and people already place light icons in their rooms — so there's no new element type, no new placement UX, and glow entities were already in the watched set.

**Overlapping pools mix.** Each is a radial gradient with `mix-blend-mode: screen` inside an isolated layer, so two lights *add* where they meet instead of the topmost winning. Isolation matters — screening against the plan itself washes out on a light theme.

The degradation ladder survives unchanged, since it was never about areas: colour-capable → its own `rgb_color`; brightness-only → `glowColor` (warm white) scaled by brightness; on/off → full strength; off/unavailable → nothing.

### What this means for your PR

**Your area work is back to exactly as you wrote it.** I reverted my changes to `areaColor`, `renderArea` and `resolveStateColor` — the `AreaPaint` refactor and the `matchStateColor` split are gone, since without a light competing for the area's fill there's nothing to rank against and the precedence trap I described no longer exists.

The only change left on your code is the `collectWatchedEntities` fix, which is still needed for `Area.entity`.

### Verification

Sampled rendered pixels with a warm lamp and a cool lamp overlapping:

| | Overlap midpoint |
| --- | --- |
| with the blend rules | `[76, 66, 76]` — both lights present, symmetric |
| without them | `[53, 60, 76]` — the topmost wins |

So the mixing is real and the CSS is what does it. **512 tests**, typecheck and build clean. The harness demo covers all four rungs, including two overlapping lamps.

Sorry for the churn on your branch — the second design is meaningfully better and it came out of your PR, so it seemed worth doing properly rather than shipping the room-wide version and revisiting later.

PR #111, opened 2026-08-03
Emit stable data-id and type classes for CSS targeting (closes #105)

### LINKED ISSUE #105: Emit a stable data-id attribute on rendered elements (areas, furniture, openings) for CSS targeting
Objects in the card config carry a stable id (area_a5r5nwl, furn_3j66s50, …), but that id never reaches the DOM. renderArea() in src/render.ts emits only three attributes:

No id, no class, no data-*, and no wrapping <g>. renderFurniture() and renderOpening() are the same — their <g> wrappers carry only a transform.

This makes the rendered SVG effectively un-targetable from CSS. Since the card lives in a normal shadow root and the <svg> is a direct child of it, card-mod can otherwise style it very cleanly — it just has nothing to select on.

Concrete use case
Area supports only static color and opacity — unlike items and furniture, which accept entity + stateColor. To drive a room's fill from a sensor (e.g. tint the living room by an illuminance reading) the only options today are:

wrap the whole card in config-template-card — which hides the card behind a wrapper and makes the excellent visual editor unreachable; or
select the polygon by its fill colour: svg polygon[fill="#62f202" i] { fill-opacity: … } — which silently breaks the moment the colour is changed in the editor.
Both are workarounds for the absence of a stable hook.

Proposed change
Emit the config id as data-id, and optionally a type class. Purely additive, no visual change, no breaking change:

(nothing from lit so the attribute is omitted rather than rendered as data-id="undefined" for objects without an id.)

The same one-liner on the <g> wrappers in renderFurniture() and renderOpening() would cover the rest.

Selectors then become stable and self-documenting:

Related idea (separate issue if preferred)
Adding entity + stateColor to Area, matching what furniture already supports, would cover the common case natively and remove the need for CSS entirely. stateColor already accepts colours with an alpha channel (both #RRGGBBAA and rgba()), so threshold-driven area transparency would work out of the box.

nicosandller:
Closes #105. Thanks @wisemars — precise issue, and the diagnosis was exactly right.

### What it does

Every rendered element now carries its config `id` as `data-id`, plus a type class:

| Element | Class | Attributes |
| --- | --- | --- |
| Area | `fp-area` | `data-id` |
| Furniture | `fp-furniture`, `fp-furniture-<type>` | `data-id`, `data-entity` |
| Door / window | `fp-opening`, `fp-opening-door` \| `fp-opening-window` | `data-id`, `data-entity` |
| Wall | `wall`, `fp-wall` | `data-id` |
| Device | `item`, `fp-item` | `data-id`, `data-entity`, `data-kind` |
| Text | `text`, `fp-text` | `data-id` |
| Tracker | `tracker`, `fp-tracker` | `data-id` |

The issue named areas, furniture and openings. I did the rest too — stopping short would leave the identical complaint for walls and devices, and it's the same one-liner each. `data-entity` is included because "style whatever is bound to `light.kitchen`" is the rule people actually want to write, and it's the one hook that survives renaming an element.

```yaml
card_mod:
  style: |
    .fp-area[data-id="area_a5r5nwl"] { fill: #62f202; fill-opacity: 0.35; }
    .fp-furniture-sofa { opacity: 0.5; }
    [data-entity="light.kitchen"] { filter: drop-shadow(0 0 6px gold); }
```

### On sanitising

Values pass through two small guards. `cssIdent` keeps `[A-Za-z0-9_-]`; `cssEntityId` also keeps the dot, since entity ids are `domain.object_id` and `[data-entity="light.kitchen"]` is what people will write — stripping the dot would make the hook silently never match. They're deliberately two functions: a dot is fine in an attribute value and wrong in a class name.

To be clear about the threat model — this **isn't** about breakout, since Lit escapes attribute values. It's about the emitted DOM being *predictable*, which is the whole point of a styling hook: whitespace in a value would quietly add a second class (a furniture type of `sofa fp-wall` would restyle every wall). Anything that sanitises to empty emits **no attribute**, per your note — `data-id="undefined"` would be a hook that matches every element lacking an id.

### Verification

On a rendered card, all seven element types resolve by `[data-id="…"]`, `data-entity` resolves the bound entity, and the markup contains no `"undefined"`.

I also checked the use case rather than just the attributes — injecting a card-mod-style block:

| | Before | After |
| --- | --- | --- |
| `.fp-area[data-id=…]` fill | `rgb(217, 202, 254)` | `rgb(0, 255, 0)` |
| fill-opacity | `0.25` | `0.75` |

CSS beats SVG presentation attributes, so `fill` / `fill-opacity` are drivable this way — the thing that made your colour-matching workaround necessary.

**538 tests** (11 new, including that a hostile id collapses to one harmless token and that a missing id omits the attribute), typecheck and build clean.

### Your "related idea" — already shipped

> Adding `entity` + `stateColor` to Area, matching what furniture already supports…

That landed in **v1.1.0**, with the editor UI for it in **v1.1.2**. Areas now take `entity`, `stateColor`, `activeColor`, `activeOpacity` and `highlight` (fill / border / both), all reachable from the visual editor — so a sensor-driven room fill needs no CSS at all now. Worth an upgrade before reaching for card-mod.

One caveat noted in the README: these are a *styling* surface, not an API. The class names are stable, but the SVG structure inside an element may change between releases, so prefer selecting the element over its internals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,7 +917,7 @@ editor.
 
 | Element | Class | Attributes |
 | --- | --- | --- |
-| Area | `fp-area` | `data-id` |
+| Area | `fp-area` | `data-id`, `data-entity` |
 | Furniture | `fp-furniture`, `fp-furniture-<type>` | `data-id`, `data-entity` |
 | Door / window | `fp-opening`, `fp-opening-door` \| `fp-opening-window` | `data-id`, `data-entity` |
 | Wall | `wall`, `fp-wall` | `data-id` |
--- a/README.md
+++ b/README.md
@@ -908,6 +908,47 @@ trackers:
       presence: { entity: binary_sensor.living_room_presence }
 ```
 
+## Styling hooks (card-mod)
+
+Every rendered element carries its config `id` as `data-id`, plus a type class — so
+[card-mod](https://github.com/thomasloven/lovelace-card-mod) and any other CSS can target
+it by something stable, instead of by a colour that breaks the moment you change it in the
+editor.
+
+| Element | Class | Attributes |
+| --- | --- | --- |
+| Area | `fp-area` | `data-id` |
+| Furniture | `fp-furniture`, `fp-furniture-<type>` | `data-id`, `data-entity` |
+| Door / window | `fp-opening`, `fp-opening-door` \| `fp-opening-window` | `data-id`, `data-entity` |
+| Wall | `wall`, `fp-wall` | `data-id` |
+| Device | `item`, `fp-item` | `data-id`, `data-entity`, `data-kind` |
+| Text | `text`, `fp-text` | `data-id` |
+| Tracker | `tracker`, `fp-tracker` | `data-id` |
+
+Ids come from the editor (`area_a5r5nwl`, `furn_3j66s50`, …) and are stable across edits.
+An element with no id simply has no `data-id`, rather than `data-id="undefined"`.
+
+```yaml
+type: custom:easy-floorplan-card
+card_mod:
+  style: |
+    /* One specific room */
+    .fp-area[data-id="area_a5r5nwl"] { fill: #62f202; fill-opacity: 0.35; }
+    /* Every sofa on the plan */
+    .fp-furniture-sofa { opacity: 0.5; }
+    /* The element bound to one entity, whatever kind it is */
+    [data-entity="light.kitchen"] { filter: drop-shadow(0 0 6px gold); }
+```
+
+CSS wins over SVG presentation attributes, so `fill` and `fill-opacity` set this way
+override what the card draws.
+
+Two notes. **Colouring a room from a sensor no longer needs CSS** — areas take `entity`,
+`stateColor`, `activeColor` and `activeOpacity` natively; see **Area**. And these hooks are
+a *styling* surface, not an API: the class names are stable, but the SVG structure inside
+an element may change between releases, so prefer selecting the element itself over its
+internals.
+
 ## Development
 
 ```bash
--- a/src/css-safe.adversarial.test.ts
+++ b/src/css-safe.adversarial.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { cssColor } from "./css-safe";
+import { cssColor, cssIdent, cssEntityId } from "./css-safe";
 import { furnitureColor } from "./render";
 import type { Furniture } from "./types";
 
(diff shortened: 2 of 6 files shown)


nicosandller:
Self-review before merge. One real bug found and fixed on the branch; the rest holds up.

### Fixed: areas didn't answer `[data-entity]`

Openings, furniture and devices all emitted `data-entity`. Areas didn't — and areas are the case this issue was actually about.

Since #107 an area takes an `entity`, so:

```css
[data-entity="binary_sensor.kitchen_smoke"] { … }
```

would have matched the door and the sofa bound to that sensor, but **not the room**. That's the least obvious way for a styling hook to be wrong: it works, just not where you most expected it, and nothing tells you. Now one selector resolves all four, verified on a rendered card:

`polygon.fp-area`, `g.fp-furniture`, `g.fp-opening`, `div.item` — 4 matches.

### Checked and fine

- **Editor unaffected.** It shares `renderArea`/`renderFurniture`/`renderOpening`/`renderTracker`, so the classes appear there too, but hit-testing is geometric (`editor-geometry.ts`), not DOM-based — and there was no `data-id` or `dataset` usage anywhere in the codebase before this, so nothing can be reading it.
- **No class-name collisions.** Existing `fp-*` names are `fp-door-arc`, `fp-glow(s)`, `fp-icon-spin`, `fp-leaf-r`, `fp-active`; `fp-area-breathe`/`fp-area-ants` are `@keyframes` names, not classes, so `.fp-area` can't collide with them.
- **`data-entity` deliberately omitted on trackers** — a tracker binds up to four entities (two axes plus presence gates), so there's no single honest value.
- **Editor walls keep no `data-id`** — the hooks are a card-mod surface for the live card; nobody styles HA's config UI.

### Two things I'd flag rather than change

**These are a styling surface, not an API.** Class names are stable, but the SVG *inside* an element may change between releases — a rule like `.fp-furniture-sofa rect:nth-child(2)` will break silently. The README says so; worth repeating in release notes if this goes out as a documented feature.

**`data-kind` duplicates config that the editor can change.** It's genuinely useful (`[data-kind="light"]`), but a device's `kind` is editable, so a rule keyed on it can stop matching after an unrelated edit. `data-id` is the stable choice and `data-entity` the meaningful one; `data-kind` is a convenience with a sharper edge. Not worth removing — worth knowing.

**539 tests** (12 new), typecheck and build clean.

nicosandller:
@copilot's one comment was correct, and worth spelling out because the mechanism is subtle.

The test claimed the attribute is omitted, but only asserted that the flattened markup doesn't contain `"undefined"`. Lit's `nothing` is `Symbol(lit-nothing)`, which stringifies to `"Symbol(lit-nothing)"` — so the check passed on a string that never demonstrated omission, and the flattened markup was quietly full of `Symbol(lit-nothing)` noise that the other assertions in that suite were reading straight past.

Verified the mechanism before acting on it:

```
typeof nothing: symbol   String(nothing): "Symbol(lit-nothing)"
values: ['symbol:Symbol(lit-nothing)', 'symbol:Symbol(lit-nothing)', 'string:0,0 1,0 1,1', …]
```

### Fixed two things

**The suite's `flatten` helper now mirrors Lit** — a symbol renders as nothing at all, so every markup assertion in that block reflects what would actually be emitted rather than sentinel text.

**The omission test asserts the real contract**: the `data-id` / `data-entity` slots are Lit's `nothing`, located by scanning the template's static strings for the attribute name rather than by slot index (which would break the moment an attribute is added).

### Confirmed the replacement is actually a guard

The old test wasn't *entirely* vacuous — it would still have failed on a raw `undefined` — but it asserted a far weaker property than it claimed. So I checked the new one both ways rather than assuming:

| | result |
| --- | --- |
| `?? nothing` removed from `renderArea` | ❌ test fails |
| restored | ✅ passes |

And confirmed the actual DOM outcome the unit test infers, on a config whose elements carry no ids:

| element | `hasAttribute('data-id')` |
| --- | --- |
| area | `false` |
| furniture | `false` |
| opening | `false` |

No `"undefined"` and no `"Symbol("` anywhere in the rendered markup.

**539 tests**, typecheck and build clean. Ready to merge from my side.

One sequencing note that hasn't changed: this and #112 both rewrite the same `return svg\`<polygon…\`` in `renderArea` and will conflict. #112 restructures more, so it should land first — and when this one rebases, the hooks need adding to the new *border* polygon too, or the outline becomes the one element on the plan CSS can't target.

PR #112, opened 2026-08-03
Areas: draw a room's outline over the walls it traces, so highlight: border works (issue #6)

superflyer11:
Follow-on to #107. `highlight: "border"` shipped in v1.1.0 but draws nothing on most plans — including mine, which is how I found it.

### The bug

An area polygon is almost always traced along the room it encloses, so it runs down the **centerline of that room's walls**. Walls are stroked at `WALL_THICKNESS` (8) on top of the area fills, and the outline was drawn with the fill. It lands underneath the very wall it follows, and 8px of wall covers a 3px outline completely.

So `highlight: border` is inert on any plan whose areas follow its walls. It only ever showed on an area that *didn't* — a zone floating inside a room.

### The fix

The outline moves to its own pass after the walls, and colors them. That turns out to be the better visual anyway: an occupied or lit room announces itself along its own boundary instead of tinting everything inside it, which is what the mode was for in the first place.

- `renderArea` keeps the fill and never strokes.
- `renderAreaBorder` returns the outline, or `nothing` — still the default.
- The new pass sits inside the **same wall mask**, so a doorway is a gap in the outline exactly as it is a gap in the wall. Without that it would paint straight across every door and window on the plan.
- Each live outline is **clipped to its own room**. Rooms share walls, and a stroke along a polygon straddles the boundary — it paints the neighbour's face as well as its own. Unclipped, a wall between two live rooms went to whichever area sat later in `areas:`, so reordering the config silently changed what the plan said. Clipped, every room paints its own side and nothing else.

### Three judgement calls

**A live border defaults to half the wall thickness, not `3`.** It's painted on the wall it traces, and a room announcing itself wants that wall colored, not a thin stripe down the middle of it. Half, not the whole 8, because the wall is centred on the same line the polygon follows — the room only owns the inner half. Anything wider runs onto the floor, over furniture standing against that wall, and out through the doorways, the opening mask only cutting `WALL_THICKNESS + 4`. Clipping keeps the inner half of a stroke, so it's drawn at twice the width — `borderWidth` stays the width you actually *see*.

**A static `borderColor` is left unclipped and drawn as authored**, at its thinner default. It's decoration placed deliberately rather than a per-room signal, so it shouldn't be silently halved or repositioned.

**Static borders moved above the walls too, not just live ones.** Splitting them across two layers seemed worse than moving both, and the move is a no-op for every outline that is currently visible: those are on areas that don't follow a wall, where there's no wall to be drawn over. The outlines this actually relocates are the ones nobody could see.

### Verification

Rasterized and pixel-sampled, rather than trusting unit tests — this is a layering bug, and unit tests on `renderArea` are exactly what missed the `collectWatchedEntities` bug on #107.

**Does the outline show at all?** Two rooms, a doorway through room A's left wall, A occupied:

| sample | before | after | at rest |
| --- | --- | --- | --- |
| room A wall, clear of the door | `(17,17,17)` | **`(76,175,80)`** | `(17,17,17)` |
| room A doorway centre | `(191,233,252)` | `(191,233,252)` | `(191,233,252)` |
| room A interior | `(191,233,252)` | `(191,233,252)` | `(191,233,252)` |

`(17,17,17)` is the wall, `(76,175,80)` is `#4caf50`. Before, the live outline is entirely buried.

**What happens where rooms meet?** A 2×2 plan — one four-way junction, two T-junctions, a doorway — with A green, B red, C blue, D at rest:

| sample | A,B,C,D | reversed | all at rest |
| --- | --- | --- | --- |
| A\|B wall — A's half | A green | A green | wall |
| A\|B wall — B's half | B red | B red | wall |
| A\|B doorway centre | fill | fill | fill |
| 4-way corner — A quadrant | A green | A green | wall |
| 4-way corner — B quadrant | B red | B red | wall |
| 4-way corner — C quadrant | C blue | C blue | wall |
| 4-way corner — D quadrant | wall | wall | wall |
| T-junction — A's half | A green | A green | wall |
| T-junction — B's half | B red | B red | wall |
| T-junction — outer face | wall | wall | wall |
| B exterior — inner half | B red | B red | wall |
| B exterior — outer face | wall | wall | wall |

Plus the two probes that catch a border overrunning the wall it sits on: 5px inward of a doorway centre (where an overhang escapes the opening cut), and the floor either side of a shared wall. Both stay clear.

Reversing the config order changes nothing, which is the point. A four-way corner splits into four quadrants, one per room. An exterior wall colors on its inside face only, so the plan's silhouette stays intact. Everything is untouched at rest.

Profiling straight across the A|B wall, the two rooms meet exactly at the centreline with no gap and no overlap — A covers x 112–120, B covers 120–128, on a wall spanning 116–124.

**539 tests** pass (+12), `tsc --noEmit` and `vite build` clean. The `renderArea` border tests moved to a `renderAreaBorder` block and picked up the clipping and mask cases.

The editor draws the same pass in the same layer position, so what you place is what you get; with no `hass` there, only a static `borderColor` draws.

### DIFF
--- a/README.md
+++ b/README.md
@@ -781,8 +781,9 @@ trackers:
   between two rooms splits down the middle and each side reports its own room, and a
   corner where several rooms meet splits between them. An exterior wall colors on its
   inside face only, leaving the plan's silhouette intact. `borderWidth` is the width
-  you actually see on the room's own side, and defaults to the thickness of the wall
-  it covers (`8`).
+  you actually see on the room's own side, and defaults to `4` — the room's own half
+  of the wall, since the wall is centred on the line the polygon follows. Widen it and
+  the band runs past the wall onto the floor and over furniture standing against it.
 
 ```yaml
 areas:
--- a/README.md
+++ b/README.md
@@ -774,10 +774,15 @@ trackers:
   inside it, which reads better on a busy plan.
 
   An area's outline is drawn **on top of the walls it traces**, so `border` colors the
-  room's own walls rather than hiding a line underneath them. A live border therefore
-  defaults to the wall thickness (`8`); set `borderWidth` yourself for a thinner line
-  inside the wall, or a thicker one that spills past it. Doorways and windows are cut
-  out of the outline exactly as they are cut out of the wall.
+  room's own walls rather than hiding a line underneath them. Doorways and windows are
+  cut out of the outline exactly as they are cut out of the wall.
+
+  A live outline is clipped to its own room, so rooms never paint each other: a wall
+  between two rooms splits down the middle and each side reports its own room, and a
+  corner where several rooms meet splits between them. An exterior wall colors on its
+  inside face only, leaving the plan's silhouette intact. `borderWidth` is the width
+  you actually see on the room's own side, and defaults to the thickness of the wall
+  it covers (`8`).
 
 ```yaml
 areas:
--- a/README.md
+++ b/README.md
@@ -773,6 +773,12 @@ trackers:
   `border` for a room that outlines itself while occupied without tinting everything
   inside it, which reads better on a busy plan.
 
+  An area's outline is drawn **on top of the walls it traces**, so `border` colors the
+  room's own walls rather than hiding a line underneath them. A live border therefore
+  defaults to the wall thickness (`8`); set `borderWidth` yourself for a thinner line
+  inside the wall, or a thicker one that spills past it. Doorways and windows are cut
+  out of the outline exactly as they are cut out of the wall.
+
 ```yaml
 areas:
   - id: living_room
@@ -801,15 +807,15 @@ areas:
       - { x: 500, y: 900 }
       - { x: 100, y: 900 }
 
-  # Outline-only highlight: the room draws a green border while occupied and
-  # its fill never changes. activeOpacity is a fill concern, so it does not
-  # apply here.
+  # Outline-only highlight: the hall's own walls turn green while it is
+  # occupied and its fill never changes. activeOpacity is a fill concern, so it
+  # does not apply here. Without borderWidth the outline matches the wall it is
+  # painted over; 4 draws a thinner line down the middle of that wall instead.
   - id: hall
     name: Hall
     entity: binary_sensor.hall_occupancy
     activeColor: "#4caf50"
     highlight: border
-    borderWidth: 4
     points:
       - { x: 500, y: 100 }
       - { x: 900, y: 100 }
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -2742,10 +2742,14 @@ export class FloorplanCardEditor extends LitElement {
               ${renderWallMask(floor.openings, c.width, c.height, this._wallMaskId)}
               ${floor.walls.map((w) => this._renderWall(w))}
               <!-- Room outlines, same layer position as the card so what you
-                   place is what you get. No hass here, so only a static
-                   borderColor draws; a live highlight is a card-time concern. -->
+                   place is what you get. Only a static borderColor draws here,
+                   there being no hass to resolve a live color from — but the
+                   clip ids are passed anyway, so wiring a live preview in later
+                   cannot silently land on the unclipped path. -->
               <g mask=${`url(#${this._wallMaskId})`}>
-                ${(floor.areas ?? []).map((a) => renderAreaBorder(a))}
+                ${(floor.areas ?? []).map((a, i) =>
+                  renderAreaBorder(a, undefined, `${this._wallMaskId}-area-${i}`)
+                )}
               </g>
               ${repeat(
                 // Keyed by id: switching floors must create fresh DOM. Reused
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -62,6 +62,7 @@ import {
   renderFurniture,
   renderTracker,
   renderArea,
+  renderAreaBorder,
   trackerSensorReading,
   kindFromEntity,
   resolveItemIcon,
@@ -2740,6 +2741,12 @@ export class FloorplanCardEditor extends LitElement {
               ${floor.furniture.map((f) => this._renderFurnitureSel(f))}
               ${renderWallMask(floor.openings, c.width, c.height, this._wallMaskId)}
               ${floor.walls.map((w) => this._renderWall(w))}
+              <!-- Room outlines, same layer position as the card so what you
+                   place is what you get. No hass here, so only a static
+                   borderColor draws; a live highlight is a card-time concern. -->
+              <g mask=${`url(#${this._wallMaskId})`}>
+                ${(floor.areas ?? []).map((a) => renderAreaBorder(a))}
+              </g>
               ${repeat(
                 // Keyed by id: switching floors must create fresh DOM. Reused
                 // nodes would CSS-transition from the previous floor's opening
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -27,6 +27,7 @@ import {
   furnitureColor,
   renderTracker,
   renderArea,
+  renderAreaBorder,
   areaColor,
   glowPaint,
   renderGlow,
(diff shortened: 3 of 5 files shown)


nicosandller (changes requested):
Strong PR. The diagnosis is right, and the clipping insight — that an unclipped shared wall goes to whichever area sits later in `areas:`, so reordering the config silently changes the plan — is the kind of thing that would have been a miserable bug report six months from now. Verifying by pixel-sampling rather than unit tests was also exactly the right call for a layering bug.

One number needs changing before merge.

### The live border default overshoots the wall

`width` defaults to `WALL_THICKNESS` (8) and is drawn at `width * 2`, clipped to the room's inside. But the wall is *also* 8 wide, **centred on the same line** — so the room only owns 4px of it. The band therefore runs 4px past the wall's inner face onto the floor.

Profiling across a left wall at `x=200`, room interior to the right:

```
x:      196 ....... 203 204 ....... 207
wall:   ############
border:          ########################
                 ^ room's half (4px)  ^ 4px on the floor
```

Measured: wall occupies `196..203`, green band occupies `200..207`.

That has three consequences, and the third contradicts the PR's own headline claim:

**1. It paints over furniture.** The border pass runs after furniture. A counter flush against a wall loses its outer 4px:

| | profile at `x=196..214` |
| --- | --- |
| at rest | `wwwwwwwwFFFFFFFFFFF` |
| occupied, default | `wwwwGGGGGGGGFFFFFFF` |

**2. The wall reads asymmetrically** — outer half black, inner half coloured, plus a band on the floor.

**3. It leaks across doorways.** The wall mask cuts holes `WALL_THICKNESS + 4 = 12` tall, i.e. ±6 from the centreline. The band runs 0..+8. So 2px sits *outside* the cut and paints straight across the opening:

| | profile through the doorway centre, `x=194..212` |
| --- | --- |
| default (8) | `............GG.....` |
| `borderWidth: 4` | `...................` |

This is worth flagging specifically because the PR says *"a doorway is a gap in the outline exactly as it is a gap in the wall"* — and it very nearly is. Your doorway sample was taken at the **centre** (`x=200`), which is masked, so it passed. The sliver is 6px inward, where a centre-only probe can't see it.

### The fix

Default the live width to `WALL_THICKNESS / 2`, not `WALL_THICKNESS`:

```diff
-  const width = cssNumber(a.borderWidth, liveBorder ? WALL_THICKNESS : DEFAULT_AREA_BORDER_WIDTH);
+  const width = cssNumber(a.borderWidth, liveBorder ? WALL_THICKNESS / 2 : DEFAULT_AREA_BORDER_WIDTH);
```

Verified: the band becomes exactly `200..203` — the room's own half of the wall, nothing on the floor, furniture untouched, and the doorway completely clean. It also makes *"painted on the wall it traces"* literally true, which the current default only half is.

Your reasoning that `borderWidth` should stay "the width you actually see" still holds — it's just that the width worth seeing is the half-wall the room owns.

### Smaller notes, none blocking

- **The editor passes no `clipId`**, which is safe only because it never passes a `liveColor` either. That's an implicit coupling: if anyone later wires live colours into the editor preview, they silently get the unclipped, config-order-dependent behaviour back — the exact bug this PR fixes. Worth either a comment at the call site or making the live path require a clip.
- **`renderArea` still emits `stroke="none" stroke-width="0"`** now that it never strokes. Harmless and keeps the DOM stable; noting it so it doesn't read as an oversight later.
- **Merge order with #111.** Both rewrite the same `return svg\`<polygon…\`` in `renderArea`, so they will conflict. Merge **this one first** — it restructures more. Then #111 rebases and must add its `data-id` / `data-entity` hooks to the *border* polygon too, otherwise the outline is the one element on the plan that CSS can't target.

Gate confirmed on my side: 538 tests, `tsc --noEmit` and `vite build` clean.

Happy to push the one-line change plus a doorway-leak regression test to your branch if you'd rather not round-trip — the fork has `maintainerCanModify`. Say the word.

superflyer11:
You're right on all three counts, and the doorway one is the part I actually got wrong rather than just tuned badly — the PR claimed openings were clean and they weren't. Pushed in 328ef13.

I reproduced each consequence before changing anything, on a wall centred at `x=200` with the interior to the right, furniture flush against its inner face, and a doorway through it. `w`=wall, `G`=border, `F`=furniture, `.`=floor:

```
x                194......201......208...213
plain wall         wwwwGGGGGGGG......   before
                   wwwwGGGG..........   after
furniture at wall  wwwwGGGGGGGGFFFFFF   before
                   wwwwGGGGFFFFFFFFFF   after
doorway centre         ......GG......   before
                       ..............   after
```

Your measurement matches mine exactly: the wall occupies `196..203`, the old band `200..207`. The `GG` at `206..207` is the sliver outside the 12-wide cut, and it's 6px inward of the opening centre — which is precisely why my centre-only probe passed. Sampling one point and calling a geometric property verified was the weak link, not the number.

`WALL_THICKNESS / 2` it is. The band is now `200..203` exactly.

What I checked after the change, since a narrower default could plausibly have broken the clipping work:

| | A,B,C,D | reversed |
| --- | --- | --- |
| A\|B wall — A's half | A green | A green |
| A\|B wall — B's half | B red | B red |
| A\|B doorway centre | fill | fill |
| A\|B doorway, 5px inward *(the old leak)* | fill | fill |
| 4-way corner — A / B / C quadrants | green / red / blue | green / red / blue |
| 4-way corner — D quadrant (at rest) | wall | wall |
| floor either side of the A\|B wall | fill | fill |

Rooms still meet exactly at the centreline, now with no spill into either interior, and order-independence is intact.

On your other two notes:

- **Editor coupling** — agreed, and it's the sharper version of the same class of bug: something that works only because two unrelated things happen to both be absent. The editor now passes clip ids regardless. It resolves no live colour so nothing changes today, but a live preview wired in later lands on the clipped path by construction rather than by luck.
- **`stroke="none"`** — kept, now with a comment saying it's deliberate: the element is reused across live/rest updates, and pinning it means the fill pass can't draw an outline whatever the border pass is doing.

**Merge order** — happy to go first, and thanks for flagging it. Worth noting for #111 that the border polygon is now *two* elements needing hooks in the live case, the `<clipPath>` and the `<polygon>`, and only the polygon should carry `data-id` — a `clipPath` isn't rendered, so a CSS rule matching it would look like it silently does nothing. I'll rebase and confirm once it lands if that's easier than you carrying it.

Gate: **539 tests** (added one tying the live default to the opening cut, so the leak can't come back), `tsc --noEmit` and `vite build` clean.

superflyer11:
Rebased onto `cb1f6bb` — conflicts resolved, back to **MERGEABLE**. Force-pushed as `e34e272`.

#111 rewrote the same `renderArea` return this PR splits in two, so git resolved it badly in a way worth naming: it kept my fill-only polygon and **silently dropped `class="fp-area"`, `data-id` and `data-entity`** from it, while stranding #111's whole return statement inside the body of my new `renderAreaBorder`. Compiles-and-conflicts is fine; compiles-and-quietly-loses-your-feature is the one to watch for. Both sides are back.

Then the substantive part, which is the thing I flagged in my last comment: the border pass is a **second polygon per area**, so without hooks the room outline would be the one element on the plan CSS can't reach — and it's the element most worth reaching, being the one that actually changes with state. It now carries `data-id` and `data-entity` exactly as the fill does, under its own `fp-area-border` class so a rule can target a room's outline and its fill separately.

```css
.fp-area-border[data-entity="binary_sensor.hall_occupancy"] { stroke-dasharray: 6 4; }
```

The hooks go on the drawn polygon **only**, never on the `<clipPath>` a live border also emits. A clipPath paints nothing, so a rule matching one looks like it silently does nothing — a miserable thing to debug. There's a test asserting exactly one `data-id` in the clipped markup so that can't drift.

Re-ran the pixel verification against the rebased tree rather than assuming #111 was orthogonal, since it touched these exact elements:

| | A,B,C,D | reversed | at rest |
| --- | --- | --- | --- |
| A\|B wall — A's half | A green | A green | wall |
| A\|B wall — B's half | B red | B red | wall |
| A\|B doorway centre | fill | fill | fill |
| A\|B doorway, 5px inward | fill | fill | fill |
| 4-way corner — A / B / C quadrants | green / red / blue | green / red / blue | wall |
| 4-way corner — D quadrant (at rest) | wall | wall | wall |
| floor either side of the wall | fill | fill | fill |

Unchanged from before the rebase: rooms meet at the centreline, corners split per room, openings clean, order-independent.

**553 tests** pass, `tsc --noEmit` and `vite build` clean. Only the `CHANGES_REQUESTED` review is holding it now.

nicosandller (approved):
Approving. You'd already done everything I was about to push, and made the same call on the one subtle bit.

I'd started resolving the conflict against #111 on your branch, but you rebased and pushed `e34e272` while I was working — so I dropped mine rather than clobber yours. Verified your version instead:

### The conflict resolution is right

Both area elements now answer the same `data-id`, with distinct classes:

`polygon.fp-area`, `polygon.fp-area-border`

And critically, the outline is `fp-area-border` and **not** also `fp-area`. That's the trap in this split: `.fp-area { fill: … }` is the rule the README documents, and the outline is drawn `fill="none"` — so had the outline carried `fp-area` too, every documented fill rule would have flooded it solid. Confirmed with that exact rule applied:

| element | computed `fill` |
| --- | --- |
| `.fp-area` | `rgb(255, 0, 255)` ✅ |
| `.fp-area-border` | `none` ✅ |

### Your own fixes survive the rebase

| profile at `x=194..212` (wall `196..203`) | |
| --- | --- |
| wall row | `..wwwwGGGG.........` |
| doorway row | `...................` |

The band is exactly `200..203` — the room's own half-wall, nothing on the floor — and the doorway is completely clean. Both symptoms I flagged are gone, and the width fix does the work of all three.

### Gate

553 tests, `tsc --noEmit` and `vite build` clean. `renderArea` and `renderAreaBorder` each defined exactly once — no silent duplication from the rebase, which is the usual way a resolution like this goes wrong.

Merges cleanly with `main` as it stands. Good to go — and thanks for the fast turnaround on the border width, the reasoning in that commit message is clearer than what I'd suggested.

PR #114, opened 2026-08-04
README: document the area outline as a styling hook, and its fill trap

superflyer11:
Doc follow-up to #112, which split an area into two rendered elements. Four small things, no code changes.

### The outline is missing from the hooks table

`polygon.fp-area-border` carries the same `data-id` / `data-entity` as the fill, and it is arguably the more useful of the two to target — it is the element that changes with state. The table listed Area as `fp-area` alone, so there was no way to discover it short of reading the DOM.

### An area is now two elements, and one of them is `fill="none"`

Worth stating outright rather than leaving to be discovered. Both polygons answer the same `data-id`, and the section already notes that CSS beats SVG presentation attributes — so the obvious rule floods the outline solid:

```css
[data-id="area_hall"]          { fill: #62f202; }  /* also floods the outline */
.fp-area[data-id="area_hall"]  { fill: #62f202; }  /* the fill, as intended */
```

You flagged exactly this when you approved #112 — that `.fp-area { fill: … }` is the rule the README documents, and giving the outline that class too would have flooded it. The class split avoids it inside the card; this makes sure a user writing their own rule doesn't walk into it.

The existing `[data-entity="light.kitchen"] { filter: … }` example stays unscoped on purpose — `filter`, `opacity` and `stroke` are fine on both — and now says so, rather than looking like an inconsistency.

### `borderWidth` had two documented defaults, twelve lines apart

`3` for a static `borderColor`, `4` for a live outline, with nothing connecting them. Each now points at the other. My omission from #112.

### Stale feature bullet

The Areas bullet still said binding an entity makes "its fill" go live.

Only `README.md` changes, so the gate is unaffected — 557 tests, `tsc --noEmit` and `vite build` clean on the branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,9 @@ automatically to the card and screen size.
   piano, hot tub.
 - **Areas** — trace a room's outline point-by-point (points snap to wall corners and to
   neighboring areas' corners) to get a colored, named room polygon. Bind an entity to a
-  room and its fill goes live — green while a presence sensor is occupied, red when a CO2
-  sensor crosses a threshold. Naming a room after
+  room and it goes live — green while a presence sensor is occupied, red when a CO2
+  sensor crosses a threshold — as a fill, or as the room's own walls lighting up if you
+  would rather not tint everything inside it. Naming a room after
   one of your Home Assistant areas (the name field autocompletes against them) links the
   two, which scopes the entity picker — for any device dropped inside it — to that HA
   area's entities, and can bulk-add every device in the area.
@@ -788,7 +789,8 @@ trackers:
 - `activeOpacity` — fill opacity while `entity` resolves a color. Lets a room lift out of
   the plan while it is live without permanently darkening it. Falls back to `opacity`.
 - `borderColor` / `borderWidth` — a static outline for the room. No outline is drawn by
-  default; `borderWidth` defaults to `3` canvas units once a color is set.
+  default; `borderWidth` defaults to `3` canvas units once a color is set. (A *live*
+  outline — see `highlight` — has its own default of `4`, for the reason given below.)
 - `highlight` — where a live color paints: `fill` (default), `border`, or `both`. Use
   `border` for a room that outlines itself while occupied without tinting everything
   inside it, which reads better on a busy plan.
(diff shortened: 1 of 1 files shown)


nicosandller (approved):
Approving. Every claim in here checks out, which for a docs PR is the whole ballgame — I verified them against the code rather than reading them.

### Verified, not taken on trust

The central warning is real. On a live `highlight: border` area, applying each rule and reading computed `fill`:

| rule | `.fp-area` | `.fp-area-border` |
| --- | --- | --- |
| *(none)* | `rgb(38,198,218)` | `none` |
| `[data-id="area_hall"] { fill }` | `rgb(98,242,2)` | **`rgb(98,242,2)`** ← flooded |
| `.fp-area[data-id="area_hall"] { fill }` | `rgb(98,242,2)` | `none` ✅ |

The `.fp-area-border { stroke-dasharray: 6 4 }` example works as written, both polygons carry `data-id` and `data-entity`, and the `3` vs `4` defaults match `renderAreaBorder` (`DEFAULT_AREA_BORDER_WIDTH` vs `WALL_THICKNESS / 2`). The "fill under the walls, outline over them" phrasing matches the actual layer order in the card.

Also confirmed the premise: main's table still reads `| Area | fp-area | … |` with no outline row, so this gap is real and not already closed by #112.

### One thing I'd add while you're in here

`stroke-width` behaves differently on the two kinds of outline, and this PR is exactly the section where someone would look it up. Measured visible width on the plan:

| | authored | visible |
| --- | --- | --- |
| live outline, default | — | **4px** ✅ matches the docs |
| live outline, `.fp-area-border { stroke-width: 20 }` | 20 | **10px** |
| static outline, `borderWidth: 10` | 10 | 10px |
| static outline, `.fp-area-border { stroke-width: 20 }` | 20 | 20px |

A **live** outline is clipped to its own room and therefore drawn at twice the width, so a CSS `stroke-width` lands at half what you asked for. A **static** one isn't clipped and behaves normally. Your prose already explains this for the *config* field — "`borderWidth` is the width you actually see" — but a card-mod user setting `stroke-width` directly gets no such warning, and the DOM shows `stroke-width="8"` where the docs say 4, which is the kind of thing that reads as a bug.

A sentence in the new CSS block would cover it, something like: *a live outline is clipped to its own room and drawn at twice the width, so a `stroke-width` you set in CSS shows at half its value; a static `borderColor` outline is not clipped and behaves normally.*

Not a blocker — the PR is strictly better than what's on main either way — so merge as-is if you'd rather, and I'm happy to add it separately.

### Note on merge order

This branches from `a475aac`, twelve commits back; main is at `6908cd7`. The README has since gained the whole **Follow the sun** section immediately above **Styling hooks**. GitHub reports mergeable and the hunks don't overlap, but the hooks table is now touched by two lines of history, so worth landing before anything else edits that table.

PR #116, opened 2026-08-04
Card: fit the plan to the card, instead of cutting it off (closes #115)

### LINKED ISSUE #115: Card is cut off, not scaled, when it is shorter than the plan's aspect ratio
### What happens

Make a card shorter than its plan's natural height — drag its bottom edge in a sections view, or set `grid_options.rows` — and the plan is **cut off** rather than scaled down. Devices in the lost strip disappear entirely: not shrunk, not clipped at the edge of an icon, simply not on screen.

Make it taller and the plan sits at the top with dead space beneath it.

### Why

`.stage` is `width: 100%` with an inline `aspect-ratio: width / height`, so its height is derived from its width alone. It never consults the height the card was actually given. `ha-card` is `height: 100%; overflow: hidden`, so anything past that height is simply clipped away.

This is already known and deliberately guarded — [`floorplan-card.ts:379-392`](https://github.com/nicosandller/easy-floorplan/blob/main/src/floorplan-card.ts#L379-L392) explains why `preserveAspectRatio="none"` is right and warns against "fixing" it, because `meet` letterboxes the SVG while the `.items` HTML overlay does not, drifting every icon off the wall it was placed on. The comment names the real fix: *"wrap the svg and the overlay in one aspect-ratio box and centre it."* This issue is just to record the symptom and the measurements.

### Reproduction

Rendering the card's DOM and CSS in headless Chromium at three card sizes, with a 1000×600 plan. A green frame is drawn at the SVG's own edges, a magenta marker at plan coordinates `(900, 550)`, and a cyan HTML overlay dot at the same logical point (`left: 90%; top: 91.7%`).

| card box | plan frame drawn | ratio | markers |
| --- | --- | --- | --- |
| 600×360 *(matches 1000×600)* | 603×363 | 1.661 ✅ | aligned, drift `(0, 0)` |
| **600×200** *(short)* | **603×203** | **2.97 ❌** | **both off-screen** |
| 600×500 *(tall)* | 603×363 | 1.661 ✅ | aligned, plan top-anchored |
| 600×auto *(masonry)* | 603×363 | 1.661 ✅ | aligned, drift `(0, 0)` |

Row two is the bug: the frame is still drawn 363 tall but only 203 of it survives, and the device at `(900, 550)` is gone. Note the alignment column — the current design's one virtue is that SVG and overlay never drift, which any fix has to preserve.

### The obvious fix does not work

Wrapping both layers in a ratio box and clamping it with `max-height: 100%` fails: when `max-height` binds, the width stays at `100%` and the ratio breaks instead of the box shrinking. Measured — it reproduces the 2.97 ratio exactly.

Sizing the box off the container height (`width: min(100%, calc(100cqh * w / h))` with `container-type: size`) does letterbox correctly at every definite height — but it **collapses the card to nothing** in a content-sized masonry view, because size containment leaves an auto height with nothing to resolve against:

| | 600×200 | 600×auto (masonry) |
| --- | --- | --- |
| `max-height` clamp | ratio 2.97 ❌ | fine |
| `container-type: size` | 336×203, ratio 1.655 ✅ | **plan not rendered** ❌ |

### What does work

Give the stage an `aspect-ratio` of its own *as well as* `height: 100%`, so it still has a definite height when the card is content-sized and size containment has something to resolve against. Then letterbox an inner ratio box against the container height and centre it:

| card box | plan frame | ratio | drift |
| --- | --- | --- | --- |
| 600×360 | 603×363 | 1.661 ✅ | `(0, 0)` |
| 600×200 | 336×203 | 1.655 ✅ | `(0, 0)` |
| 600×500 | 603×363 | 1.661 ✅ | `(0, 0)` |
| 600×auto | 603×363 | 1.661 ✅ | `(0, 0)` |

Whole plan shown at every size, centred, and the overlay stays pinned to the SVG — so `preserveAspectRatio="none"` remains correct, the box always matching the viewBox ratio. PR to follow.

superflyer11:
Closes #115.

`.stage` is `width: 100%` with an `aspect-ratio`, so its height comes from its width alone and never consults the height the card was actually given. `ha-card` is `height: 100%; overflow: hidden`, so a card shorter than that has the bottom of its plan cut away — devices down there don't shrink or clip at the edge of an icon, they are simply not on screen.

### The fix

The `<svg>` and the `.items` overlay now sit in one plan box that carries the canvas ratio, fitted into the card and centred — so the two layers letterbox **together**. That is what the comment above the `<svg>` has been asking for, and it leaves `preserveAspectRatio="none"` correct on its own terms: the box always matches the viewBox ratio, so `"none"` still never distorts.

### Two things that don't work, for the record

**Clamping a full-width box with `max-height: 100%`.** When `max-height` binds, the width stays at `100%` and the *ratio* breaks rather than the box shrinking. Measured — it reproduces the bug's own wrong ratio exactly. So the box is sized off the container height instead.

**That needs `container-type: size`, which has its own trap.** Size containment means an auto-height stage has nothing for `100cqh` to resolve against, and in a content-sized masonry card the plan **collapsed to nothing** — a far worse regression than the clipping it fixes. Giving the stage the canvas ratio *as well as* `height: 100%` leaves it with a definite height either way. Both dead ends are in the commit message so nobody re-derives them.

### Verification

I drove the **built card** in headless Chromium rather than a hand-written mock, since this is entirely a layout bug and a mock proves nothing about the real DOM. The canvas background is set to a flat colour so the green region *is* the plan box; an area polygon covers plan coords `(800,500)-(1000,600)` in the SVG layer, and a text label sits at `(900,550)` in the HTML overlay.

| card box | before | after |
| --- | --- | --- |
| 600×360 *(matches the plan)* | `600×360` · 1.667 · aligned | `600×360` · 1.667 · aligned |
| **600×200** *(short)* | `600×200` · **3.0** · **both markers gone** | `334×200` · **1.67** · aligned |
| 600×500 *(tall)* | `600×360` · 1.667 · aligned | `600×360` · 1.667 · aligned |
| 600×auto *(masonry)* | `600×360` · 1.667 · aligned | `600×360` · 1.667 · aligned |

Only the short card changes. In every case the overlay marker lands inside its own SVG area — the property the existing comment exists to protect, and the one I'd expect you to check hardest.

The floor switcher stays pinned to the stage rather than the plan box, being card chrome rather than part of the drawing. The canvas background moved onto the plan box, so it marks the plan exactly and letterbox bars fall back to the card surface; in an exactly-fitting card the two are the same element, so nothing changes there.

**557 tests** pass, `tsc --noEmit` and `vite build` clean.

### DIFF
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -373,9 +373,25 @@ export class FloorplanCard extends LitElement {
       <ha-card .header=${c.title ?? nothing}>
         <div
           class="stage"
-          style="aspect-ratio: ${dims.w} / ${dims.h}; background:${cssColorOr(
-          c.background, "var(--card-background-color, #fff)")};"
+          style="aspect-ratio: ${dims.w} / ${dims.h};"
         >
+          <!-- The plan box: exactly the canvas ratio, fitted inside whatever
+               height the card was actually given, and centred there (closes
+               #115). Sized off the container's height so it shrinks when the
+               height is the binding axis — clamping a full-width box with
+               max-height instead would break the ratio rather than the box.
+
+               The stage carries the same aspect-ratio so it still has a
+               definite height in a content-sized (masonry) card; without it,
+               size containment leaves 100cqh with nothing to resolve against
+               and the plan collapses to nothing. -->
+          <div
+            class="plan"
+            style="aspect-ratio: ${dims.w} / ${dims.h};
+                   width: min(100%, calc(100cqh * ${dims.w} / ${dims.h}));
+                   background:${cssColorOr(
+                     c.background, "var(--card-background-color, #fff)")};"
+          >
 <!-- preserveAspectRatio="none" is correct here, and it took a wrong fix to
                see why. .stage pins aspect-ratio: width / height inline, so the
                SVG's box already matches its viewBox and "none" never distorts.
@@ -522,6 +538,7 @@ export class FloorplanCard extends LitElement {
               (it) => this._renderItem(it, c, rot)
             )}
           </div>
+          </div>
           ${floors.length > 1 ? this._renderFloorSwitcher(floors, active) : nothing}
         </div>
       </ha-card>
@@ -564,7 +581,18 @@ export class FloorplanCard extends LitElement {
     .stage {
       position: relative;
       width: 100%;
+      height: 100%;
       padding: 0;
+      /* Centres the plan box in whatever the card was given, and makes the
+         stage's own height queryable so the plan can size against it. */
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      container-type: size;
+    }
+    .plan {
+      position: relative;
+      height: auto;
     }
     .floor-switcher {
       position: absolute;
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -440,20 +440,19 @@ export class FloorplanCard extends LitElement {
                    background:${cssColorOr(
                      c.background, "var(--card-background-color, #fff)")};"
           >
-<!-- preserveAspectRatio="none" is correct here, and it took a wrong fix to
-               see why. .stage pins aspect-ratio: width / height inline, so the
-               SVG's box already matches its viewBox and "none" never distorts.
+          <!-- preserveAspectRatio="none" is correct here, and it took a wrong
+               fix to see why. Fitting the plan into a card that is the wrong
+               shape for it is .plan's job, not this line's (#115): .plan
+               carries the canvas ratio, so the SVG's box always matches its
+               viewBox, and "none" and "meet" are equivalent while that holds.
 
-               "meet" letterboxes the SVG inside its box. The .items overlay is
-               HTML, positioned with raw left/top percentages of .stage, and it
-               does not letterbox. So the moment anything overrides the stage's
-               ratio (card-mod, a grid row count) the drawing shrinks away from
-               the badges and every icon drifts off the wall it was placed on.
-               "none" stretches both layers identically: distorted, but aligned.
-
-               The real fix letterboxes both layers together -- wrap the svg and
-               the overlay in one aspect-ratio box and centre it. Until then, do
-               not "fix" this line. -->
+               "none" is still the deliberate choice, because it is the one
+               that fails safely. The .items overlay is HTML, positioned with
+               raw left/top percentages of .plan, and it does not letterbox. So
+               if anything ever overrides .plan's ratio (card-mod, a grid row
+               count), "meet" letterboxes the SVG away from the overlay and
+               every icon drifts off the wall it was placed on, while "none"
+               stretches both layers identically: distorted, but aligned. -->
           <svg viewBox="0 0 ${dims.w} ${dims.h}" preserveAspectRatio="none">
             <g transform=${rotTransform || nothing}>
             ${active.image
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -577,11 +577,21 @@ export class FloorplanCard extends LitElement {
       height: 100%;
       box-sizing: border-box;
       overflow: hidden;
+      /* A column, so the stage takes the height left over after the card's
+         own header rather than the card's whole height. With a title set, a
+         full-height stage measures past the bottom of the card by exactly the
+         header, and the plan is cut off by that much. */
+      display: flex;
+      flex-direction: column;
     }
     .stage {
       position: relative;
       width: 100%;
-      height: 100%;
+      /* Takes the space the header leaves, and may shrink below its content:
+         without min-height a flex item floors at its content size and the
+         plan pushes the stage past the card again. */
+      flex: 1 1 auto;
+      min-height: 0;
       padding: 0;
       /* Centres the plan box in whatever the card was given, and makes the
          stage's own height queryable so the plan can size against it. */


superflyer11:
Went back over this looking for edge cases and found one the PR missed, plus one behaviour change worth calling out explicitly. Pushed as `d0e82bb`.

### A card with a `title` was still clipped

The plan box is fitted to the stage, but the stage was `height: 100%` of the whole card — *including* the space the header occupies. So with a title the stage measured past the bottom of the card by exactly the header height, and the plan lost that much. Same clipping as #115, one element further in.

It predates #116 rather than being caused by it; the first version of this PR just didn't fix it. `ha-card` is now a flex column so the stage takes what's left after the header, with `min-height: 0` so it can shrink below its content — a flex item floors at its content size by default, which pushes the plan straight back past the card.

**My earlier verification of this case was worthless and I'd rather say so than have you find it.** `ha-card` is undefined in a bare browser, so `.header` was set on an element that renders nothing — the title scenarios were measuring a card with no header and passing trivially. Re-run against a stand-in `ha-card` that actually renders one:

| scenario | card | plan drawn | ratio | want |
| title, natural ratio | 800×480 | `800×424` → `706×424` | 1.887 → **1.665** | 1.667 |
| title, short card | 800×300 | `500×244` → `406×244` | 2.049 → **1.664** | 1.667 |

### Eleven scenarios now hold

plain and titled × natural / short / tall / narrow, `rotation: 90` at both natural and short, two floors with the switcher, masonry auto-height, and title-in-masonry.

### One deliberate behaviour change

A card **taller** than its plan used to top-anchor the drawing and leave the dead space below it; it now centres the plan in that space. Measured: plan at `y=0` before, `y=110` after in an 800×700 card. That falls out of "fit and centre" and I think it's the better default, but it is a visual change on cards that were never broken, so it should be your call rather than something you discover.

The floor switcher deliberately stays pinned to the stage rather than the plan box, being card chrome rather than part of the drawing — so on a letterboxed card it sits in the card's corner, not the plan's.

**557 tests**, `tsc --noEmit` and `vite build` clean.

nicosandller (changes requested):
The fix is right and the investigation behind it is excellent. One thing must change before merge, and it's a comment rather than code — which sounds minor and isn't, for reasons below.

I verified independently rather than trusting the table, driving the built card at four sizes:

| card | plan box | ratio | inside card | device at (900,550) |
| 600×360 natural | 597×358 | 1.667 | ✅ | ✅ visible, aligned |
| **600×200 short** | **330×198** | **1.667** | ✅ | ✅ visible, aligned |
| 600×500 tall | 598×359 | 1.667 | ✅ | ✅ visible, aligned |
| 300×360 narrow | 298×179 | 1.667 | ✅ | ✅ visible, aligned |

The SVG matches the plan box to sub-pixel in every case, and the overlay marker lands where the SVG puts it — the property that comment exists to protect.

### The comment is now stale, and self-defeating

It still reads:

> `.stage` pins aspect-ratio: width / height inline, so the SVG's box already matches its viewBox…
> The .items overlay is HTML, positioned with raw left/top percentages of **.stage**…
> **The real fix letterboxes both layers together — wrap the svg and the overlay in one aspect-ratio box and centre it. Until then, do not "fix" this line.**

**This PR is that fix.** It wraps both layers in `.plan`, gives it the canvas ratio, and centres it. So the comment now names the wrong element twice, and — worse — leaves a standing instruction whose stated precondition has silently been satisfied. The next person to read it can't tell whether the warning still binds, which is the one thing a landmine comment must never do.

The conclusion survives; the reasoning changes. `none` stays correct, but no longer because the overlay can't letterbox — now because the box always matches the viewBox, so `none` and `meet` are equivalent *and* `none` remains the defensive choice: if card-mod ever overrides `.plan`'s ratio, `none` stretches both layers identically (distorted but aligned) where `meet` would letterbox the SVG away from the overlay. That's worth writing down, since it's the actual justification going forward.

### The centring change — my call, as asked

**Keep it.** "Fit and centre" is what people expect of a letterbox, and top-anchoring only ever looked deliberate because the plan happened to start at the top. It does change cards that were never broken, so it belongs in the release notes rather than being discovered.

### Merge state

This branches from `a475aac`, six PRs back — main is now at `6908cd7`. GitHub says mergeable, but that's textual, so I merged locally and re-ran: clean, **577 tests**, `tsc --noEmit` and `vite build` both fine.

Specifically checked the sun dimming that landed after you branched: its overlay rect sits inside the SVG in plan coordinates, so it follows the new box for free. Its bounding rect reports 5.3px wider than the plan box, which had me looking — that's exactly the intended `WALL_THICKNESS` padding, and it's geometric only: the SVG computes `overflow: hidden`, so nothing spills into the letterbox bars. Confirmed visually on a 600×200 card.

### Credit where it's due

Catching that your own `ha-card` verification was worthless — because the stub renders nothing, so every title scenario was silently testing a card with no header — and saying so unprompted is the most valuable thing in this PR. Both documented dead ends (`max-height` breaking the ratio, `container-type: size` collapsing masonry) will save whoever touches this next a real afternoon.

### One judgement call I'd note but not block on

The floor switcher stays pinned to `.stage`, so on a heavily letterboxed card it sits in the card's corner with a gap between it and the plan it controls. You flagged it as deliberate and the reasoning holds (it's chrome, not drawing). I'd lean the other way — it's a control *for the plan* — but it's defensible and reversible, so it shouldn't hold this up.

superflyer11:
Comment rewritten, and you were right that it isn't minor — a landmine whose precondition has quietly been met is worse than no comment, because the next reader has to re-derive whether it still binds before they can touch the line.

Pushed as `659a01b`.

### The new comment

```
preserveAspectRatio="none" is correct here, and it took a wrong
fix to see why. Fitting the plan into a card that is the wrong
shape for it is .plan's job, not this line's (#115): .plan
carries the canvas ratio, so the SVG's box always matches its
viewBox, and "none" and "meet" are equivalent while that holds.

"none" is still the deliberate choice, because it is the one
that fails safely. The .items overlay is HTML, positioned with
raw left/top percentages of .plan, and it does not letterbox. So
if anything ever overrides .plan's ratio (card-mod, a grid row
count), "meet" letterboxes the SVG away from the overlay and
every icon drifts off the wall it was placed on, while "none"
stretches both layers identically: distorted, but aligned.
```

That's your reasoning rather than mine — the conclusion survives, the justification is now the one that will still be true next year, and there's no standing order left to expire. Both `.stage` references are gone.

### I did not force-push over your merge

Your `8f5734b` was on the branch when I went to push. I'd rebased locally, which would have erased it, so I threw the rebase away and rebuilt on top of your merge instead: `8f5734b` → merge of current `main` → the comment commit. It's a fast-forward, and your commit is still in the history.

Worth knowing that `main` has moved five commits past where you tested — it's at `c5842d3` now (the device badge dropdown and the glow reach change landed after your review). So I re-verified there rather than asking you to trust the earlier run.

### Re-verified on c5842d3

Twelve scenarios, and this time in **both** the measured DOM and the painted pixels — the plan box is a flat colour, so the green region *is* the plan and I can check what actually reached the screen rather than what `getBoundingClientRect` claims:

| | painted plan | ratio | SVG≡plan | overlay marker | bleed outside card |
| --- | --- | --- | --- | --- | --- |
| plain natural / short / tall / narrow | matches DOM ±0.7px | 1.667 | ✅ | ✅ on the SVG marker | none |
| title natural / short / tall | matches DOM ±0.7px | 1.667 | ✅ | ✅ | none |
| rotation 90, natural / short | matches DOM | 0.600 | ✅ | ✅ | none |
| two floors, masonry, title-in-masonry | matches DOM | 1.667 | ✅ | ✅ | none |

And the bug still reproduces without the branch, on current `main`:

| short card | before (`c5842d3`) | after |
| --- | --- | --- |
| plain 600×200 | plan drawn `600×360`, **overlay marker 0 painted px** | `333×200` · 1.667 · marker painted |
| title 800×300 | plan drawn `800×480`, **0 px** | `373×224` · 1.667 · painted |
| rot90 600×400 | plan drawn `600×1000`, **0 px** | `240×400` · 0.600 · painted |
| floors 600×200 | plan drawn `600×360`, **0 px** | `333×200` · 1.667 · painted |

Zero painted pixels is the bug stated precisely: the device isn't small or clipped at an icon edge, it was never on screen.

### One thing I chased and it's clean — but the reason is thin

Making `ha-card` a flex column means the header and the stage are now flex *siblings*, so a card too short for its plan overflows and flexbox shrinks **both** items. I first measured the header squashing from 56px to 50px, and to 31px on a short card.

That turned out to be my stub, not the card — I'd given the stub header an explicit `height`, and a definite height is exactly what lets `min-height: auto` clamp below it. HA's real `.card-header` sets no height at all; its size comes from `line-height: 48px` plus 12/16 padding, so its automatic minimum *is* its natural size and it cannot shrink. Re-measured against a byte-for-byte copy of HA's rule: **76px in every scenario, including the ones that overflow.**

So there's nothing to fix, but I'd rather you know the guarantee is "HA's header has no explicit height" than have it read as designed-in. If HA ever sets one, the header shrinks before the plan does. I couldn't defend against it from here anyway — the header lives in `ha-card`'s own shadow root and exposes no part.

Two harness bugs of my own, since last time was the same shape: the stub `ha-card` had no background, because I'd styled it from a document rule and page stylesheets don't reach into a shadow root — so every "letterbox bar" I measured was actually the page showing through. And `requestAnimationFrame` doesn't fire under `--virtual-time-budget`, so the first run measured nothing at all and reported it as a pass.

### Your two calls

**Centring — kept**, as you decided. It's yours for the release notes.

**Floor switcher** — left pinned to `.stage`. You said it shouldn't hold this up, so I haven't pre-empted you, but I'd rather move it than have it be the thing you notice on a wall tablet. Say the word and it's a one-line change to hang it off `.plan`.

**641 tests**, `tsc --noEmit` and `vite build` clean on the merged tree.

nicosandller (approved):

Issue #108, opened 2026-08-03
Editor draws every glow light at full strength even when the light is off


nicosandller:
Since v1.1.0. **Editor only — the live card is correct.**

### Symptom

With `glow: true` on several lights, the editor canvas washes warm amber over the plan whether or not those lights are on. Everything under the pools — furniture especially, since it fills at 0.12 opacity — reads as though it were coloured or "active", even with no conditional colouring configured anywhere.

It's worst on a plan with many glow lights in one room, which is the normal case: a living room with five lamps paints five full-strength pools on top of each other.

### Cause

`src/editor.ts` falls back to a full-strength pool whenever `glowPaint` declines:

```ts
const paint = glowPaint(it, this.hass?.states[it.entity]) ?? {
  color: cssColorOr(it.glowColor, DEFAULT_GLOW_COLOR),
  opacity: GLOW_MAX_OPACITY,
};
return renderGlow(it, paint, ...);
```

The intent was a preview when there's no `hass` at all, so the radius stays adjustable without turning the real light on. But it's applied unconditionally — `glowPaint` also returns `undefined` for a light that is simply **off**, and that case falls into the same fallback. So an off light previews identically to a fully-on one.

The card has no such fallback, which is why it renders correctly.

### Reproduction

Five living-room lights, all `off`, all `glow: true`:

| | Glow circles drawn |
| --- | --- |
| editor | **5**, each `#ffd9a0` at `stop-opacity 0.6` |
| card | 0 ✅ |

### Fix

Trust the entity state when there is one: an off light draws nothing, matching the card. Keep the preview only for a glow with no readable state (no `hass`, or an entity `hass` doesn't know).

That alone would make radius editing hard on an off light, so the editor should draw a **dashed radius guide** for the selected glow item instead — editor-only chrome, in the same spirit as the tracker zone outline. Sizing stays easy, without lying about what's lit.

nicosandller:
Reopening the scope: my first read of this was wrong, and the reported symptom was a **different, larger bug**. There are three, and they share one cause.

I fixed an editor-only preview bug first (off lights previewing lit). That was real, but it wasn't what was reported — @nicosandller was seeing yellow furniture on the **live card**. I'd sampled while every glow light happened to be `off`, so the card looked clean and I stopped too early. With a lamp actually on, it reproduces immediately.

### 1. Furniture takes the color of the light *(the reported bug)*

[...]

Issue #102, opened 2026-08-01
When rotating display to 90 or 270, walls disappear beyond a certain point.


thewan056:
When rotating using 90 or 270, beyond a certain point walls disappear. Only walls as far as i can tell. It works fine if left at 0 or 180.

<img height="600" alt="Image" src="https://github.com/user-attachments/assets/020bec3e-2763-4f4f-852d-f358ff1401ab" />

PR #118, opened 2026-08-04
Follow the real sun: dim the plan at dusk, brighten it at dawn (closes #113)

### LINKED ISSUE #113: Add option to follow natural sun lighting
Add an optional to darken the floorplan gradually as the sun sets and brighten the floorplan gradually as the sun rises.

We can add the option to set max brightness and min brightness if the option is enabled. Light should follow the local time of the HA instance and the sunrise and sunset time for that day on the HA instance (not browser)

nicosandller:
Closes #113.

```yaml
sunDimming: true
sunBrightnessMin: 0.45   # once the sun is fully down
sunBrightnessMax: 1      # full daylight
```

Toggle it in the editor under **Project → Follow the sun**; the two sliders appear once it's on.

### Why `elevation` and not sunrise/sunset times

The issue asks for the **HA instance's** local time and sun times, not the browser's. `sun.sun` already carries an `elevation` attribute that Home Assistant computes continuously from the instance's latitude, longitude and clock — so using it settles that requirement by construction rather than by careful timezone handling. It's also already smooth, so there's nothing to interpolate.

A phone in another timezone showing the same dashboard sees the same picture.

The ramp spans civil twilight, **−6° to +6°** — roughly the hour around sunrise and sunset when the light outside actually changes — smoothstepped so the rate eases in and out instead of cornering at each end.

### Where the dim lands

One rect, last inside the rotated group. It covers the plan; the HTML device overlay sits above it and stays at full brightness, so a night plan reads as a dark house with legible icons and its lit rooms still glowing:

| sun elevation | black overlay opacity |
| --- | --- |
| noon (+60°) | `0` |
| day (+6°) | `0` |
| golden (+3°) | `0.086` |
| sunset (0°) | `0.275` |
| dusk (−3°) | `0.464` |
| civil dusk (−6°) | `0.55` |
| night (−30°) | `0.55` |
| night, `sunBrightnessMin: 0.15` | `0.85` |
| option off | no rect at all |

`pointer-events: none`, because a canvas-spanning rect that eats taps is exactly how #108 broke area selection.

### Two traps this had to avoid

**Watched entities.** `sun.sun` joins the watched set when the option is on. Without it the plan would be lit once and then frozen — the same trap entity-bound furniture (#82) and areas (#6) each fell into. Confirmed the card repaints on **pure hass ticks** where only the attribute moves: elevation 20° → 3° → 1° all report state `above_horizon`, and each one repaints.

**Failing bright is harder than it looks.** `Number(null)`, `Number("")`, `Number(false)` and `Number([])` are *every one of them* `0` — and 0° is the **middle** of this ramp, not night. A naive `Number.isFinite` guard would have left a dead `sun.sun` sitting at half light, looking like a dusk that never ends. My own test caught it mid-implementation; the input is now allowlisted to numbers and non-empty strings, and there's a test for each coercion.

### Deliberate choices worth a second opinion

- **`sunBrightnessMin` defaults to `0.45`, not `0`.** A plan you can't read at night is worse than one that's merely dim.
- **Cast-light pools dim with the plan.** They still stand out against darkened rooms (see the screenshot case in the description), but they're not exempt. Exempting them would mean reordering the glow layer above furniture, which would undo the "light cast on the floor" layering from #109 — not worth it for this. Raise `sunBrightnessMin` if you want them to blaze.
- **The editor canvas doesn't dim.** It's an editing surface; dimming it would make night editing worse for no benefit. The card preview in the editor dialog *is* the card, so it does dim.

**567 tests** (13 new), `tsc --noEmit` and `vite build` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -976,8 +976,27 @@ A few deliberate choices:
   than one that is merely dim. Set it lower if you want a darker house.
 - **It fails bright.** If `sun.sun` is missing or unreadable the plan stays at full
   brightness, rather than being stranded dark with nothing on screen to explain why.
-- **Cast-light pools dim with the plan.** They still stand out against the darkened rooms,
-  but if you want them to blaze at night, raise `sunBrightnessMin`.
+
+### Lit rooms hold back the night
+
+A flat dim would darken a lit room as much as an empty one — multiplying the whole picture,
+contrast included — so a lamp would end up *less* noticeable at night than at noon. Instead,
+**light withholds the dim**: any device with **Cast light** on clears the darkness around
+itself, full at the centre and diffusing to nothing at its `glowRadius`, the same shape and
+falloff as the pool it casts.
+
+The result is a dark house with genuinely lit rooms. Measured against the same spot in an
+unlit room, a lamp's contrast goes from **100 by day to 116 at night** on a dark theme, and
+from 25 to 106 on a light one — where a flat dim would have cut both to 45%.
+
+Strength follows brightness the way the pool does: a full-brightness lamp clears completely,
+one dimmed to nothing clears about a third. A light that is off, `unavailable`, or has no
+Cast light enabled clears nothing.
+
+One limitation: the clearing is a plain circle, so it bleeds a little through a wall into the
+next room — measured at roughly 6% of the lit room's brightness just past the wall, fading to
+nothing within about 80 units. The pools themselves already stop at walls; the clearing does
+not yet.
 
 Toggle it in the editor under **Project → Follow the sun**; the two brightness sliders
 appear once it is on.
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ automatically to the card and screen size.
   An optional occupancy `binary_sensor` per axis gates the animation so the
   marker hides cleanly when the room is empty. The zone outline is visible only
   in the editor — the live card shows just the animation.
+- **Follow the sun** — optionally dim the whole plan through dusk and brighten it through
+  dawn, tracking your Home Assistant's own sunrise and sunset rather than the viewer's
+  browser. Device icons stay at full brightness, so a night plan reads as a dark house
+  with its lit rooms glowing.
 - **Text labels** and a configurable **canvas background color**.
 - **Background image** — drop in a floor-plan image (per floor) and trace walls, doors and
   devices over it, with adjustable opacity.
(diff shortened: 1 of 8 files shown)


nicosandller:
Answering the question on strategy — and the honest answer was that the original approach got this backwards, so it's now fixed on the branch.

### What was wrong

The dim was one flat black rect over everything. A flat overlay multiplies the *whole* picture, contrast included — so it dimmed the gap between a lit room and an unlit one just as much as it dimmed each of them. Measured, sampling a lamp's pool against the same spot in a room with no lamp:

| | day | night | |
| --- | --- | --- | --- |
| Dark theme | +100.1 | **+45.1** | |
| Light theme | −25.5 | **−11.4** | |

Night contrast was exactly **45%** of daytime in both themes — which is `1 − dimOpacity`, the dim factor itself. A lamp was *less* prominent after dark than at noon. My PR description said the pools "still stand out"; that was wrong, and only measuring showed it.

### What it does now

Rather than dimming over the light, **the light withholds the dim**. The dim rect carries a mask built from the same radial falloff the pools already use: full clearing at a lamp's centre, diffusing to full dim at its `glowRadius` — so the clearing and the pool are the same shape by construction, exactly as you asked.

| | day | night | |
| --- | --- | --- | --- |
| Dark theme | +100.1 | **+116.1** | lit 130 held, unlit 30.9 → 14 |
| Light theme | +25.5 | **+106.3** | lit 215 held, unlit → 108.7 |

Worth noting the light-theme row: previously the warm pool was slightly *darker* than the near-white floor around it, so a lamp never read as a light source at all. At night it now does.

Only **Cast light** devices clear the dark — they're the ones that define a radius — and strength tracks brightness the way the glow does: full-brightness clears completely, a lamp dimmed to nothing clears about a third, off/unavailable clears nothing.

### Verification

- The mask **states its own region**, so rotation can't clip the clearing: it stays `1016×616` in plan coordinates while the viewport swaps to `600×1000` at 90°. That default is what clipped walls in #102.
- `.fp-sun-dim` still computes `pointer-events: none`, and an opening under a lit pool keeps its hit target (#108).
- 574 tests (7 new), `tsc --noEmit` and `vite build` clean.

### One limitation, measured not guessed

The clearing is a plain circle, so it bleeds through walls. Profiling across a wall from a lamp 110 units away with a 200 radius:

| distance past the wall | luminance (baseline 14) |
| --- | --- |
| +10 | 20.9 |
| +40 | 17.9 |
| +80 | 14.8 |
| +120 | 14 |

So ~6% of the lit room's brightness just past the wall, gone within about 80 units. Visible if you look for it, not obtrusive. The *pools* already stop at walls via `glowReach` (#109); reusing it to clip the clearing is the clean follow-up, and I'd rather do it as its own change than widen this one — say the word if you'd prefer it in here.

PR #125, opened 2026-08-05
Devices: icon per state rule, and the reading inside the badge (closes #106)

### LINKED ISSUE #106: Replace Icon with Text/Change Icon
I have a thermostat and have already color-coded it according to the temperature. I also display the status below it.

Is it also possible to display the temperature (rounded) instead of the icon? 

I also have blinds, and I'd like to assign different icons to "open" and "closed." Is that possible?

nicosandller:
Closes #106.

@MrMcFlyy asked for two things, with a [mockup](https://github.com/nicosandller/easy-floorplan/issues/106#issuecomment-5177230125): a thermostat badge with the temperature *in* it rather than a text line underneath, and different icons for blinds open vs. closed. Their own summary was the right framing — *"you can not only change the color, but also the icon depending on the state."*

Both land here, and between them they add **one** row to the device form, which was the constraint: that panel is already crowded.

## 1. Icon per state rule

`StateColorRule` gains an optional `icon`. The "Color by state" list people already use becomes "Color & icon by state", with one more box per row:

```yaml
stateColor:
  - state: open
    color: "#4caf50"
    icon: mdi:blinds-open
  - state: closed
    color: "#9e9e9e"
    icon: mdi:blinds
```

No new panel, no new concept, and it works for any device — a lock, a washing machine, a sensor above a threshold — not just covers.

`resolveStateColor` is now a one-line wrapper over a new `matchStateRule`, so the colour and the icon always come off the **same** matched rule. Resolving the precedence separately for each would be two chances to drift, and would quietly allow one rule's colour beside another rule's icon.

A rule's icon sits **above** the config `icon`, for the same reason its colour already beats `activeColor`: it is the more specific statement about right now. That also undoes a trap worth calling out — setting a config `icon` used to `return` early in `resolveItemIcon`, freezing the glyph and silently disabling every state-dependent icon below it. A custom icon and state-dependent icons are now usable together.

## 2. The reading inside the badge

`badgeContent: icon | value | none` replaces the `showIcon` boolean, which is deprecated and still honoured when unset — so every existing config renders identically, and in the form it is the *same row*, now a three-way dropdown.

`badgeValue()` picks the reading per domain and stops at the first **numeric** candidate:

1. the configured `attribute`
2. the domain's default reading (`climate` → `current_temperature`, `humidifier` → `current_humidity`)
3. the entity's own state
4. the secondary entity's reading

That numeric gate is what makes step 1 safe to put first, and it handles the mockup's own case: colouring a thermostat by `attribute: hvac_action` gives `"heating"` — text — so it falls through and the badge still shows the temperature. **You can colour by one reading and display another with no extra config.**

Step 4 is what makes a smart plug work: a `switch` with `secondaryEntity: sensor.plug_power` reads `1.2kW` and still toggles the switch on tap. And when nothing numeric turns up anywhere, the badge keeps its icon — this can never leave an empty circle.

Deliberately *not* routed through `hass.formatEntityState`: that applies display precision and the full unit (`21.5 °C`), which is exactly what does not fit in a 34px circle.

### Two things worth flagging

**Sizing measures glyphs, not characters.** My first pass sized by string length and it was wrong — `45%` is wider than `9999`, and `21°` is narrower than `782`. It put `1240W` 3.2px *outside* an 18px badge. The widths are now measured off the rendered card and rounded up, and the parity nudge goes *down* where `itemIconSize` goes up, since this size has just been clamped to a width budget. Remaining known limit: a 5-glyph reading in an 18px badge hits the 6px legibility floor and grazes the rim by 0.4px — below that it would be a smudge rather than a number, so it wants a bigger badge.

**`cssIcon` validates rather than strips**, unlike its `cssIdent` neighbour. My own test caught why: stripping `"><script>` leaves `script` — harmless in the DOM, but it has now displaced the icon the caller would otherwise have fallen back to. Callers rely on `undefined` to move to the next candidate, so a value that cannot render must not survive as *something*.

`collectWatchedEntities` needs no change — `entity` and `secondaryEntity` were already watched and climate attributes ride on the primary state object. Flagging it explicitly because a miss there paints once and freezes (it bit #82, #6 and #113).

## Verification

- `tsc --noEmit`, 609 tests, `npm run build` all green; duplication sweep clean.
- Every new regression test was **run against `main` first** — 26 of them fail there; the two that don't are deliberate no-change guards.
- Dev harness, 15 fixtures: values, icons, colours and font sizes **identical** between the card and the editor canvas (the `_renderItemOverlay` duplication is the standing trap here).
- Measured rendered text width vs. badge width at 18 / 24 / 34 / 80px.
- Rotation 90°: badge stays `rotate(0deg)` so the number is upright, while an item with its own `angle: 45` keeps it.
- Legacy `showIcon: false` still renders label-only; items still clickable.

Also fixed in passing: the README's `kind` row still listed only the original 7 kinds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,9 +222,42 @@ By default it shows an icon badge:
   makes a wall of lights, covers and switches hard to read. Set **Active color**
   per device to tell them apart at a glance — lights yellow, covers purple,
   thermostats orange. Ripples follow it unless you set a ripple color too.
-- **Color by state** — the **Color by state** rules color the element by what the
-  entity reads: both the **icon badge** and the label, whether or not the entity is
-  "on" (a temperature sensor never is). Add rules in the editor, or in YAML:
+- **Value in the badge** — set **Badge shows** to *Value* and the device draws its
+  reading inside the badge instead of an icon: a thermostat reads `21°` in the same
+  circle your state rules already paint red while it's heating, with no text line
+  hanging underneath. The card picks the reading per domain, so there is nothing to
+  configure — a climate shows its current temperature (its *state* is the mode,
+  "heat"), a humidifier its humidity, a sensor its own value. Long units are dropped
+  to keep the number legible: `°C` becomes `°`, CO₂ reads plainly as `780`.
+
+  An **Attribute** you set wins when it's a number, so you can colour by one reading
+  and display another — `attribute: hvac_action` colours the badge by whether the
+  boiler is firing while the badge still shows the temperature. For a smart plug, point
+  **2nd entity** at its power sensor: the badge reads `1.2kW` and tapping still toggles
+  the switch. A device with no number anywhere keeps its icon, so this can never leave
+  an empty circle. *Nothing* is the third setting — no badge at all, label only.
+- **Color and icon by state** — the **Color & icon by state** rules restyle the element
+  by what the entity reads: both the **icon badge** and the label, whether or not the
+  entity is "on" (a temperature sensor never is). Each rule can also swap the **icon**,
+  so blinds get an open glyph and a closed one:
+
+  ```yaml
+  stateColor:
+    - state: open
+      color: "#4caf50"
+      icon: mdi:blinds-open
+    - state: closed
+      color: "#9e9e9e"
+      icon: mdi:blinds
+  ```
+
+  `icon` is optional — a rule without one only changes the colour, exactly as before.
+  A matching rule's icon beats the device's fixed **icon** override, which is what makes
+  a custom icon and state-dependent icons usable together; previously setting one froze
+  the glyph for good. (Covers that carry a **device class** already get open/closed icons
+  with no rules at all — this is for everything else, and for a third state.)
+
+  Add rules in the editor, or in YAML:
 
   ```yaml
   stateColor:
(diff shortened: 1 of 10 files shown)


nicosandller:
Pushed a second commit (71b32d7) picking up the three reports from the issue thread. The linked review was Copilot's overview, which generated no comments, so nothing there needed action.

### Lights badge themselves — @ombre33

A bulb that reports an `rgb_color` now wears it, darkened in step with brightness. Precedence is **state rules → `activeColor` → the bulb's own colour → theme yellow**, so anything set by hand still wins, and the ripple follows the same chain.

New `lightBadgePaint`, deliberately **not** a reuse of `glowPaint` despite how similar they look: that one falls back to a warm white so a pool always has something to cast, and reusing it would have repainted every plain on/off bulb amber on installs that never asked for it. Verified: a brightness-only and an on/off bulb both stay theme-yellow.

Brightness scales the channels rather than the alpha — a translucent badge composites against the *plan* behind it, so the same lamp would read differently over a dark room than a light one.

### A readable glyph — @MrMcFlyy

Reproduced the report exactly. With a dark theme and a white state colour, the badge computed `rgb(255,255,255)` on `rgb(255,255,255)` — invisible. The background comes from config while the ink was pinned to `--text-primary-color`; the two are independent, so the ink now follows the background.

`contrastText` compares the two actual WCAG contrast ratios rather than testing a luminance threshold. A threshold has to be derived from the inks, and the usual 0.179 crossover assumes pure black where ours is `#212121` — which moves it to ~0.212. My first pass used 0.179 and a test caught it picking the *worse* ink on part of the grey ramp. Comparing ratios can't drift.

It returns `undefined` for anything it can't resolve — `var()`, `color-mix()`, gradients — and those badges keep the theme ink exactly as before.

Measured on the live card, every badge in the fixture: **minimum ratio 5.21**, all ≥ WCAG AA.

### Light lands on furniture — @MrMcFlyy

This one is a **re-balance, not a revert**, and worth being explicit about: furniture was cut out of the glow layer on purpose, as the fix for #108 where translucent furniture let a pool shine through and every sofa read as highlighted. Cutting it out entirely overcorrected into a shadow — a lit table came out *darker* than the floor beside it, which is what was reported here.

Both complaints are ends of one dial, so the mask is a grey now rather than `black`. Sampling a table and bare floor at **equal distance from the lamp** (the radial falloff makes unequal distances incomparable — my first measurement made exactly that mistake), colour cast measured as R−B:

| transmission | table | floor | vs. unlit table | |
|---|---|---|---|---|
| **0** (before) | **0** | 76 | 0 | invisible — and a *cold* patch beside warm floor |
| **0.5** (now) | **33** | 76 | **−34** | clearly lit, still clearly furniture |
| **1.0** (#108) | **66** | 76 | −67 | essentially flooded |

### Editor parity, and a gap it exposed

The canvas previewed **none** of the active colours — setting "Active color" changed nothing there. Previewing only the new automatic colour would have been actively misleading (you'd see a bulb's colour but not the explicit setting that overrides it), so the editor now mirrors the card's whole chain. That fixes a pre-existing gap as a side effect.

Verified across nine fixtures: **zero mismatches** in computed background/foreground between card and editor, including an off light.

### Checks

- `tsc`, **621 tests**, `npm run build` green; duplication sweep clean.
- The 12 new tests were run against the previous commit's source first — all 12 fail there.
- `collectWatchedEntities` needs no change: `hassRenderInputsChanged` compares state-object identity and HA replaces that object when any attribute moves, so colour and brightness changes already re-render. Confirmed rather than assumed — this is the trap that bit #82, #6 and #113.

PR #138, opened 2026-08-06
Walls: per-wall adjustable thickness

obbers:
## Summary
- Adds an optional `thickness` field to `Wall` (defaults to `WALL_THICKNESS` when unset)
- Adds a "Thickness" slider to the wall property panel in the editor, following the same default/strip pattern as other size fields (e.g. item size)
- Both the editor canvas and the live card now render each wall at its own thickness

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` (641 tests passing, including new coverage for the thickness field default/patch behavior)
- [x] `npm run build`
- [x] Manually verified in `npm run serve`: selecting a wall shows the Thickness slider, and dragging it live-updates the wall's stroke width in both the editor and the preview pane

### DIFF
--- a/src/editor-forms.test.ts
+++ b/src/editor-forms.test.ts
@@ -533,6 +533,18 @@ describe("wallForm / projectForm / floorImageForm", () => {
     expect(d).toMatchObject({ x1: 1, y1: 3, x2: 3, y2: 4 });
   });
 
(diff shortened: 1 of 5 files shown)


obbers:
Closing per request.

PR #148, opened 2026-08-09
Overlay: size badges and labels in canvas units so they scale with the plan

superflyer11:
### The problem

The card draws in two layers and only one of them scales.

Walls, doors, furniture and room fills are SVG mapped from the canvas to whatever width the card gets. Badges, labels, room names and text are HTML on top of that — they have to be, so they stay upright under `rotation` and can take clicks — and every measure in that layer is a **screen pixel**: `.area-label` is a hard-coded `font-size: 14px`, an item label defaults to 12px, `.badge` takes `width:${size}px` straight from config.

The two layers only agree while the card renders at roughly its canvas size. Below that they come apart. A 980×700 plan shown 500 wide draws every wall at half size while a 14px room name stays 14px — so the name is effectively twice as wide as the room it names, spilling past its walls and landing on the badges under it.

Nothing in the config fixes it. A label's px size has no idea what scale the plan ended up at, so "make it smaller" is guesswork that breaks again at the next card width. And an area's name had no size option at all — `showName: false` was the only answer to a name wider than its room.

### The change

`overlayScale: plan` sizes the overlay in **canvas units**, the same units as everything else in the config, so the two layers scale together and the card reads the same at any width.

```yaml
width: 980
height: 700
overlayScale: plan
```

Every config-driven overlay measure follows: badge `size` and its icon, the reading drawn inside a badge (#106), item `labelSize`, text `size`, `rippleSize`, and the new area `labelSize`.

Mechanically `.plan` becomes an `inline-size` query container and `--fp-u` is one canvas unit as a length, so `calc(14 * var(--fp-u))` is 14 canvas units however large the card ends up. Two things about that are deliberate and commented in the CSS:

- **The container is `.plan`, not `.stage`.** Since #115 the stage is only the box the plan is *centred in*, and it is wider than the plan on any card that isn't the canvas's ratio. Measuring the stage would oversize every label by exactly the letterboxing.
- **`--fp-u` is declared on `.items`, inside `.plan`, never on `.plan` itself.** Container units resolve against an element's nearest *ancestor* container, so `100cqw` read on `.plan` would measure the stage. (`.plan`'s own `width` reads `100cqh` and keeps resolving against `.stage`, for the same reason — adding `inline-size` containment to `.plan` doesn't disturb it.)

Hairlines deliberately do **not** scale: a badge border and a label's drop shadow are about a pixel either way, and scaling them with the plan puts them below a pixel on exactly the small cards this mode is for. Skins own those tokens now in any case.

`fixed` stays the default, so no existing card changes appearance on upgrade, and a wall tablet showing a plan at full size keeps px sizing, which is what keeps text legible from across the room.

### Also

- **`Area.labelSize`** (8–40, default 14 — the value it replaces). Rooms differ in size and their names differ in length; a shared hard-coded 14 meant the smallest room set the ceiling for everyone.
- Editor: the mode sits with **Rotate display** in the bottom Project row (both are set-once display choices, and like rotation it affects the live card, not the editor canvas). Area gains a **Name size** row, shown only while the name renders — same rule the item form already uses for its label size.

### Verified

Unit tests can't see this one, so it was checked against a real built card at two shapes. With the card at the canvas ratio and at 900×300 — where #115's letterbox makes `.plan` about 40 % narrower than `.stage` — a room name measured **11.06 %** and **11.03 %** of the plan's wall-to-wall width. Constant, i.e. the overlay tracks the drawing. Had it measured `.stage`, the second would have been ~23 % larger.

Config-driven sizes still go through `cssNumber` and their clamps before reaching `overlayLength`, so what lands in the inline style is a number, never a config string; covered by tests alongside the existing `itemLabelSize` injection cases. 744 tests pass. README documents the option, the area field, and when each mode is the right one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,7 @@ The editor writes this config for you; manual editing is optional.
 | `sunBrightnessMax` | number | `1` | Brightness in full daylight, 0–1. |
 | `skin`       | string   | `default`          | Built-in look for the whole plan: `default`, `odnetnin`, `pastel` or `tron`. See [Skins](#skins). |
 | `pressEffect`| string   | `scale`            | Feedback when a device is pressed: `scale`, `ripple`, `flash` or `none`. Only devices that actually do something respond. See [Press feedback](#press-feedback). |
+| `overlayScale`| string  | `fixed`            | How badges, labels, room names and text are sized: `fixed` = screen pixels, `plan` = canvas units so they scale with the drawing. See [Overlay scale](#overlay-scale). |
 | `background` | string   | skin / card bg     | Canvas background color (CSS / hex). Overrides the skin's paper. |
 | `floors`     | Floor[]  | —                  | Per-floor element groups (see [Floor](#floor)).   |
 | `defaultFloor`| string  | first floor        | Id of the floor shown first.                 |
@@ -511,11 +512,13 @@ animated inside a rectangular tracked area:
 
 ### Area
 
-`{ id, points, name?, showName?, color?, opacity?, haArea?, filterEntities?, entity?, stateColor?, activeColor?, activeOpacity?, borderColor?, borderWidth?, highlight? }`
+`{ id, points, name?, showName?, labelSize?, color?, opacity?, haArea?, filterEntities?, entity?, stateColor?, activeColor?, activeOpacity?, borderColor?, borderWidth?, highlight? }`
 
 - `points` — `{ x, y }` vertices in drawing order, implicitly closed last-to-first.
 - `name` / `showName` — label centered on the polygon (`showName` defaults `true`).
   Mirrors the linked HA area's name when `haArea` is set.
+- `labelSize` — that label's size, `8`–`40`, default `14`. Px under `overlayScale: fixed`,
+  canvas units under `plan`. Small rooms want a smaller number than the big ones beside them.
 - `color` / `opacity` — the room's fill; theme primary and `0.25` by default.
 - `haArea` — id of a linked Home Assistant area, set by the editor when `name` matches one.
 - `filterEntities` — with `haArea` set, scopes the entity picker for devices inside this
@@ -783,6 +786,42 @@ card_mod:
       --fp-skin-accent: #f2aa4c !important;
     }
 ```
+
+## Overlay scale
+
+The card draws in two layers. Walls, doors, furniture and room fills are SVG, scaled from
+the canvas to whatever width the card gets — draw at any size, they always fit. Badges,
+labels, room names and text are HTML on top of that, so they stay upright under
+`rotation` and can take clicks, and by default they are sized in **screen pixels**.
+
+Those two agree while the card renders at roughly its canvas size. Below that they drift
+apart: a `980`-wide plan shown `500` wide draws every wall at half size while a 14px room
+name stays 14px, so names spill past their rooms and collide with the badges under them.
+Nothing in the config can fix that, because a label's px size doesn't know what scale the
+plan ended up at.
+
+`overlayScale: plan` sizes the overlay in **canvas units** instead, so both layers shrink
+together and the card looks the same at every size — a scale drawing rather than a drawing
+with fixed-size furniture on it. Every measure follows: `size` and `labelSize` on a
+device, the reading drawn inside a badge, `size` on text, an area's `labelSize`, and
+`rippleSize`. Hairlines deliberately don't — a badge border and a label's drop shadow
+are about a pixel either way, and scaling them down is how you lose them.
+
+```yaml
+type: custom:easy-floorplan-card
+width: 980
+height: 700
+overlayScale: plan
+```
+
+Use it whenever the card renders smaller than its canvas — a dashboard tile, a sidebar, a
+desktop widget. Leave it off for a wall tablet showing the plan at full size, where fixed
+px is what keeps text legible from across the room. The default stays `fixed`, so an
+existing card looks exactly as it did.
+
+Sizes are read in canvas units under `plan`, so the numbers mean the same thing as
+everything else in the config: `labelSize: 14` is 14 units on a `980`-unit-wide canvas,
+about 1.4 % of the card's width whatever that turns out to be.
 
 ## Styling hooks (card-mod)
 
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ animated inside a rectangular tracked area:
   Mirrors the linked HA area's name when `haArea` is set.
 - `labelSize` — that label's size, `8`–`40`, default `14`. Px under `overlayScale: fixed`,
   canvas units under `plan`. Small rooms want a smaller number than the big ones beside them.
+  Left unset on a `fixed` card the size stays in the stylesheet, so a card-mod rule on
+  `.area-label` still wins; set it, or switch to `plan`, and it moves inline and takes over.
 - `color` / `opacity` — the room's fill; theme primary and `0.25` by default.
 - `haArea` — id of a linked Home Assistant area, set by the editor when `name` matches one.
 - `filterEntities` — with `haArea` set, scopes the entity picker for devices inside this
@@ -814,14 +816,35 @@ height: 700
 overlayScale: plan
 ```
 
-Use it whenever the card renders smaller than its canvas — a dashboard tile, a sidebar, a
-desktop widget. Leave it off for a wall tablet showing the plan at full size, where fixed
-px is what keeps text legible from across the room. The default stays `fixed`, so an
-existing card looks exactly as it did.
-
 Sizes are read in canvas units under `plan`, so the numbers mean the same thing as
 everything else in the config: `labelSize: 14` is 14 units on a `980`-unit-wide canvas,
 about 1.4 % of the card's width whatever that turns out to be.
+
+### Where it helps, and where it costs
+
+Scaling with the drawing cuts both ways: it stops text overflowing its room, and it also
+keeps shrinking past the point text can be read. On a `980`-wide canvas at the default
+sizes (room name `14`, device label `12`):
+
+| Card width | Room name | Device label |
+| --- | --- | --- |
+| 1200 | 17px | 15px |
+| 800 | 11px | 10px |
+| 600 | 9px | 7px |
+| 450 | 6px | 6px |
+| 350 | 5px | 4px |
+
+So `plan` suits a card rendering down to roughly **two-thirds of its canvas width** on the
+defaults. Below that it trades collision for illegibility, and the sizes have to come up
+to compensate — a `labelSize` of `20`–`24` on a card at half its canvas width lands back
+where the default was. That is a real trade, not a free win: sizes are relative, and
+nothing puts a floor under them.
+
+Use it whenever the card renders smaller than its canvas but not drastically so — a
+dashboard tile, a sidebar, a desktop widget. Leave it off for a wall tablet showing the
+plan at full size, where fixed px is what keeps text legible from across the room, and
+for a card so small that scaled text would disappear. The default stays `fixed`, so an
+existing card looks exactly as it did.
 
 ## Styling hooks (card-mod)
 
@@ -839,6 +862,7 @@ it by something stable.
 | Wall | `wall`, `fp-wall` | `data-id` |
 | Device | `item`, `fp-item` | `data-id`, `data-entity`, `data-kind` |
 | Text | `text`, `fp-text` | `data-id` |
+| Room name | `area-label` | — |
 | Tracker | `tracker`, `fp-tracker` | `data-id` |
 
 Ids come from the editor (`area_a5r5nwl`, `furn_3j66s50`, …) and are stable across edits.
--- a/src/editor-forms.ts
+++ b/src/editor-forms.ts
@@ -949,7 +949,7 @@ export function projectSkinForm(c: FloorplanCardConfig): FormSpec {
  * tablet's orientation, a dashboard tile's size), not day-to-day editing, so
  * they stay out of the way.
  */
-export function projectRotationForm(c: FloorplanCardConfig): FormSpec {
+export function projectDisplayForm(c: FloorplanCardConfig): FormSpec {
   return {
     fields: [
       {
--- a/src/editor-forms.ts
+++ b/src/editor-forms.ts
@@ -20,6 +20,7 @@ import type {
 } from "./types";
 import {
   DEFAULT_AREA_OPACITY,
+  DEFAULT_AREA_LABEL_SIZE,
   DEFAULT_GRID,
   DEFAULT_ITEM_SIZE,
   DEFAULT_RIPPLE_SIZE,
@@ -35,6 +36,7 @@ import {
   badgeContentOf,
   domainIconAnimation,
   isPresenceEntity,
+  normalizeOverlayScale,
   normalizePlanRotation,
   openingMotion,
   pressEffectOf,
@@ -787,6 +789,19 @@ export function areaForm(a: Area): FormSpec {
   return {
     fields: [
       { name: "showName", label: "Show name", selector: { boolean: {} } },
+      // Only while the name renders — same rule the item form uses for its
+      // label size.
+      ...((a.showName ?? true)
+        ? [
+            {
+              name: "labelSize",
+              label: "Name size",
+              selector: {
+                number: { min: 8, max: 40, step: 1, mode: "slider" as const, unit_of_measurement: "px" },
+              },
+            },
+          ]
+        : []),
       {
         name: "opacity",
         label: "Fill opacity",
@@ -827,6 +842,7 @@ export function areaForm(a: Area): FormSpec {
     ],
     data: {
       showName: a.showName ?? true,
+      labelSize: a.labelSize ?? DEFAULT_AREA_LABEL_SIZE,
       opacity: a.opacity ?? DEFAULT_AREA_OPACITY,
       entity: a.entity ?? "",
       activeOpacity: a.activeOpacity ?? a.opacity ?? DEFAULT_AREA_OPACITY,
@@ -928,9 +944,10 @@ export function projectSkinForm(c: FloorplanCardConfig): FormSpec {
 }
 
 /**
- * Display rotation (issue #33), a separate one-field form so the editor can
- * render it as the very last Project row — it's a set-once option for wall
- * tablets, not day-to-day editing, so it stays out of the way.
+ * Display-only options, a separate form so the editor can render them as the
+ * very last Project rows — set-once choices for how the card is shown (a wall
+ * tablet's orientation, a dashboard tile's size), not day-to-day editing, so
+ * they stay out of the way.
  */
 export function projectRotationForm(c: FloorplanCardConfig): FormSpec {
   return {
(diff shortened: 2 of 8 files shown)


superflyer11:
Rebased onto main — the letterbox landing in #116 changed the right answer here, so this is worth a note rather than a silent force-push.

The container moved from `.stage` to `.plan`. When this was written the stage *was* the plan's box; since #116 it is the box the plan is centred in, and it is wider than the plan whenever the card isn't the canvas's ratio. Left on `.stage`, `--fp-u` would have oversized every label by exactly the letterboxing — worst on the small, wrongly-shaped cards this mode exists for. `.plan`'s own `width: min(100%, 100cqh * w/h)` is unaffected: its `100cqh` still resolves against `.stage`, because container units always look at an element's *ancestor* container, which is the same rule that keeps `--fp-u` on `.items` rather than on `.plan`.

Also folded in since: the badge value from #106 scales with everything else, and badge borders/shadows no longer scale at all — skins own those tokens now, and a 1.5px hairline scaled to 0.8px is a hairline you've lost.

Re-verified against a real built card rather than the test suite: at the canvas ratio and at 900×300 (where `.plan` is ~40 % narrower than `.stage`), a room name measures 11.06 % and 11.03 % of the plan's wall-to-wall width. 744 tests pass.

nicosandller (commented):
## Review

The mechanism is sound, and rather than take the PR body's word on the parts unit tests can't reach, I built the branch and measured them in a browser. **Approve with changes** — one narrow regression to fix, one design gap the README currently oversells. Everything else below is a nit.

### Verified independently

- **744 tests pass, typecheck clean.** ✅
- **The container is `.plan`, not `.stage`** ✅ — at a card shape that letterboxes `.plan` to 417px inside an 898px stage, the label tracked the plan. Measuring the stage would have made it ~2.15x larger, so this was worth getting right and it is right.
- **`--fp-u` is exact** ✅ — font-size came out at precisely `14 x planW/980` at every size tested.
- **Rotation** ✅ — correct at all four angles; `dims.w` is the right divisor.
- **Containment doesn't disturb `.plan`'s own `100cqh` width** ✅ — the plan box is identical between modes (417.20 x 297.99 either way), which is the claim I most expected to find broken.
- **Injection safety** ✅ — every `overlayLength` call site passes through `cssNumber` or a clamp, and the parameter is typed `number`, so a config string can't reach the style sink.

The CSS comments explain the genuinely non-obvious parts (why `.plan` and not `.stage`, why the overlay and not the container) better than most code of this kind manages.

### Needs a change before merge

1. **Room-name card-mod overrides break** — see the inline comment on `.area-label`. Narrow, but it hits precisely the people who cared enough to work around the missing option.
2. **`plan` makes text illegible on the card sizes the README recommends it for** — measurements inline on the Overlay scale section. Not a blocker, but the docs shouldn't present it as a free win.

### Also worth knowing

The badge's 1.5px border deliberately doesn't scale. At a 417px plan that's 3px on a 17.5px badge, so badges read chunkier as the card shrinks — a larger proportional effect than "about a pixel either way" suggests. I think the call is right, since a sub-pixel border just vanishes, but the visible consequence is bigger than the comment implies.

`Area.labelSize` has no editor-canvas preview, though the editor never drew room names at all — pre-existing, not introduced here.

src/floorplan-card.ts
```diff
@@ -1145,7 +1218,8 @@ export class FloorplanCard extends LitElement {
       white-space: nowrap;
       transform: translate(-50%, -50%);
       font-weight: 600;
-      font-size: 14px;
+      /* Size is inline, from the area's labelSize (default
```
nicosandller: **Breaks card-mod overrides of room-name size — in both modes, including the `fixed` default.**

Moving this from the stylesheet to an inline style means it now beats any non-`!important` card-mod rule. Verified on this branch with `overlayScale` unset (the upgrade path):

```
inline style: "left:50%; top:50%; font-size:14px;"
inject `.area-label { font-size: 28px }` -> computed: 14px
```

Before this PR that injection won, because the size lived here as a normal rule. So "no existing card changes appearance on upgrade" holds for the default *rendering*, but not for anyone who used card-mod — and per the PR's own framing (room names had no size option), card-mod was the only prior way to change this.

Suggest emitting the inline `font-size` only when it would differ from this default — `scale === "plan"`, or `labelSize` explicitly set — and keeping the stylesheet rule otherwise. `.text` and `.label` were already inline before this PR, so `.area-label` is the only affected hook.

README.md
```diff
@@ -784,6 +787,42 @@ card_mod:
...
+are about a pixel either way, and scaling them down is how you lose them.
+
+```yaml
+type: custom:easy-floorplan-card
+width: 980
+height: 700
+overlayScale: plan
+```
+
+Use it whenever the card renders smaller than its canvas — a dashboard tile, a sidebar, a
```
nicosandller: **This recommends `plan` for exactly the sizes where it makes text illegible.**

Measured on this branch, default sizes (area name 14u, item label 12u) on a 980-wide canvas:

| Card width | Plan width | Area name | Item label |
|---|---|---|---|
| 1200 | 1198 | 17.1px | 14.7px |
| 800 | 795 | 11.4px | 9.7px |
| 600 | 598 | 8.5px | **7.3px** |
| 450 | 445 | 6.4px | **5.5px** |
| 350 | 347 | 5.0px | **4.3px** |

Below roughly 800px the labels can't be read. The collision problem is real and this does fix it, but on a small card it trades collision for illegibility, and this paragraph reads as though `plan` is strictly better there.

Two options, and I'd take the second unless you feel strongly:

1. A floor — `max(8px, calc(N * var(--fp-u)))` — friendlier, at the cost of the exact proportionality below that size.
2. Say plainly that `plan` suits cards down to about canvas-width/1.5, and that below that the sizes need raising past their defaults.

Either way the honest framing is that this is a trade, not a free win.

src/editor-forms.ts
```diff
@@ -787,6 +789,19 @@ export function areaForm(a: Area): FormSpec {
...
       { name: "showName", label: "Show name", selector: { boolean: {} } },
+      // Only while the name renders — same rule the item form uses for its
+      // label size.
+      ...((a.showName ?? true)
+        ? [
+            {
+              name: "labelSize",
+              label: "Name size",
+              selector: {
+                number: { min: 8, max: 40, step: 1, mode: "slider" as const, unit_of_measurement: "px" },
```
nicosandller, line 800: `unit_of_measurement: "px"` is wrong under `overlayScale: plan`, where this number is canvas units — which is the point of the PR.

`areaForm(a)` has no access to the config, so making it conditional needs plumbing. Six other sliders already have the same wart (`size`, `labelSize`, `rippleSize`, text `size`, …), so this is the new field inheriting an existing problem rather than creating one — probably a follow-up issue rather than something to block on. Flagging it so it's a decision and not an oversight.

src/render.ts
```diff
@@ -919,6 +921,40 @@ export function itemLabelSize(v: unknown): number {
...
+  return v === "plan" ? "plan" : "fixed";
+}
+
+/**
+ * A CSS length for one overlay measure: screen px under `fixed`, canvas units
+ * under `plan`. `units` is already coerced by the caller (cssNumber and the
+ * per-measure clamps), which is what makes this safe to drop into an inline
+ * `style` — it interpolates a number, never a config string.
+ */
+export function overlayLength(units: number, scale: OverlayScale): string {
```
nicosandller, line 959: Nit, and theoretical today: `var(--fp-u)` has no fallback, so a consumer rendering outside `.items` under `plan` makes the property invalid at computed-value time and `font-size` silently inherits. Every current call site is inside `.items`, so nothing is broken — but `var(--fp-u, 1px)` costs nothing and turns a silent wrong size into a visible-but-sane one.

src/render.test.ts
```diff
@@ -734,6 +737,37 @@ describe("editorItemLabel (#135)", () => {
   });
 });
 
+describe("overlay scaling", () => {
```
nicosandller, line 742: These cover `overlayLength` / `normalizeOverlayScale` / `areaLabelSize` well, but nothing asserts that `scale` is actually *threaded through* a render path. Dropping the argument at one of the six call sites in `_renderItem` / `_renderBadge` would still pass CI.

`renderRipple` is exported and `flattenMarkup`-testable, so one assertion anchors the wiring cheaply:

```ts
expect(flattenMarkup(renderRipple(true, "#fff", 80, 3, "plan")))
  .toContain("width:calc(80 * var(--fp-u))");
```

src/floorplan-card.ts
```diff
@@ -783,6 +819,43 @@ export class FloorplanCard extends LitElement {
       position: relative;
       height: auto;
     }
+    /*
+     * overlayScale: plan. The container is .plan, not .stage: since #115 the
+     * stage is only the box the plan is *centred in*, and it is wider than the
+     * plan on any card that isn't the canvas's ratio. Measuring the stage would
+     * oversize every label by exactly the letterboxing.
+     *
+     * --fp-u -- one canvas unit as a length -- is declared on the overlay
```
nicosandller, line 831: The placement is right but this reasoning isn't quite. `--fp-u` is an *unregistered* custom property, so its value is substituted as a token stream and the `cqw` resolves at the element that finally uses it — which is a descendant of `.plan` either way. Declaring it on `.plan` would have worked too.

Worth keeping it on `.items` regardless, since that stays correct if `--fp-u` is ever registered via `@property` (which *would* resolve it at the declaring element). So: right call, and the comment would be stronger making that the argument rather than "100cqw read on .plan would measure the stage."

src/editor-forms.ts
```diff
@@ -928,9 +944,10 @@ export function projectSkinForm(c: FloorplanCardConfig): FormSpec {
 }
 
 /**
- * Display rotation (issue #33), a separate one-field form so the editor can
- * render it as the very last Project row — it's a set-once option for wall
- * tablets, not day-to-day editing, so it stays out of the way.
+ * Display-only options, a separate form so the editor can render them as the
```
nicosandller, line 947: The doc comment got updated for the second option but the name didn't — `projectRotationForm` now owns two unrelated display settings. `projectDisplayForm` would match what it does.

nicosandller (changes requested):
## Requesting one change

Formalising the blocking item from [my review above](https://github.com/nicosandller/easy-floorplan/pull/148#pullrequestreview-4903092302) — the detail is in the inline threads, so this is just the summary and the bar for approval.

### Blocking

**[Room-name card-mod overrides break on upgrade](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374090)** — `.area-label`'s `font-size` moves from the stylesheet to an inline style in **both** modes, so it ships to everyone, including people who never set `overlayScale`. That makes the PR's "no existing card changes appearance on upgrade" inaccurate, and this is a HACS card where people upgrade without reading release notes.

Verified on the branch with `overlayScale` unset:

```
inline style: "left:50%; top:50%; font-size:14px;"
inject `.area-label { font-size: 28px }` -> computed: 14px   (was 28px before this PR)
```

Fix: emit the inline `font-size` only when it would differ from the stylesheet default — `scale === "plan"`, or `labelSize` explicitly set — and keep `.area-label { font-size: 14px }` for the untouched case. `.text` and `.label` were already inline pre-PR, so this is the only affected hook.

### Not blocking, but please decide rather than skip

**[The README recommends `plan` for card sizes where it makes text illegible](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374092)** — measurements in the thread; a 12u item label is 7.3px at a 600px card and 5.5px at 450px. Either add a `max(8px, ...)` floor or say plainly in the docs that this is a trade below roughly canvas-width/1.5. I don't mind which, but the current wording presents it as a free win.

### Follow-ups, fine to leave

The remaining four ([`px` unit](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374094), [`var()` fallback](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374095), [threading test](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374099), [`projectRotationForm` name](https://github.com/nicosandller/easy-floorplan/pull/148#discussion_r3755374109)) are all fine as separate issues. The `px` unit one predates this PR across six sliders, so it shouldn't be this PR's job.

---

To be clear about the overall shape: the engineering is good and I checked the hard parts rather than trusting the description — container-vs-stage, `--fp-u`'s exactness, all four rotations, and containment leaving the plan box untouched all held up. This is one small fix from an approve.

src/floorplan-card.ts
superflyer11: Fixed in cdcd1dd, and the finding is real — but the repro in this comment doesn't reproduce, so the fix is aimed at a slightly different rule than the one quoted.

I built three versions and injected into the card's shadow root the way card-mod does, on a default card (no `overlayScale`, no `labelSize`):

| card-mod rule | branch point | reviewed commit | now |
| --- | --- | --- | --- |
| `.area-label { font-size: 28px }` | 14px | 14px | 14px |
| `.area-label { font-size: 28px !important }` | 28px | 28px | 28px |
| `ha-card .area-label { font-size: 28px }` | **28px** | **14px** | **28px** |

Row 3 is the regression, and it's back. Row 1 — the exact rule quoted above — was already losing at 14px before this branch: lit puts `static styles` in `adoptedStyleSheets`, and the cascade orders those *after* a `<style>` injected into the same shadow root, so an equal-specificity rule loses on tree order no matter what this branch does. Pre-existing, and I've left it alone.

So the conclusion holds — anyone who out-specified the card's own rule, which is what people actually write, broke — and your suggested fix is the one I took: emit the inline `font-size` only when `scale === "plan"` or `labelSize` is set, and keep the stylesheet rule otherwise.

The decision is in `areaLabelFontSize` in [`render.ts`](https://github.com/nicosandller/easy-floorplan/blob/cdcd1dd/src/render.ts) rather than inline in the card, so it's unit-testable where `_renderAreaLabel` isn't — four cases covering the default, an explicit size, `plan`, and a hand-edited size reaching the sink.

README.md
superflyer11: Took option 2, in cdcd1dd. The section now carries your table and states the trade rather than implying there isn't one:

> So `plan` suits a card rendering down to roughly **two-thirds of its canvas width** on the defaults. Below that it trades collision for illegibility, and the sizes have to come up to compensate — a `labelSize` of `20`–`24` on a card at half its canvas width lands back where the default was. That is a real trade, not a free win: sizes are relative, and nothing puts a floor under them.

No `max(8px, …)`: a floor would make the mode silently stop being proportional at exactly the sizes someone reached for it to *be* proportional, and it'd do that without saying so. Better to make the user raise the number and know why. The "use it when…" paragraph now says "smaller than its canvas but not drastically so", and names the too-small card as a case to leave it off.

Also documented `.area-label` in the styling-hooks table and, on `Area.labelSize`, the precedence rule the fix above creates: unset on a `fixed` card the size stays in the stylesheet and card-mod still wins; set it, or switch to `plan`, and it moves inline and takes over.

src/editor-forms.ts
superflyer11: Agreed, and left as a follow-up per your read of it. Making `areaForm(a)` conditional means threading the config into it, and the same change would want to cover the six sliders already carrying the wrong suffix — that's a change to the editor's form plumbing, not to this branch.

Flagging it back as a decision rather than an oversight: the new field inherits the existing wart knowingly. Happy to open the issue covering all seven.

src/render.ts
superflyer11: Done in cdcd1dd — `calc(N * var(--fp-u, 1px))`. Costs nothing and, as you say, turns a silent inherit into something visibly wrong but sane. The doc comment on `overlayLength` now says why it's there, since every current call site makes it look like dead weight.

Confirmed in a browser that it doesn't perturb the real path: inside `.items` the computed size is still exactly `14 × planW/980`, to four decimals.

src/render.test.ts
superflyer11: Added in cdcd1dd, close to what you wrote:

```ts
expect(flattenMarkup(renderRipple(true, "#fff", 80, 3, "plan")))
  .toContain("width:calc(80 * var(--fp-u, 1px))");
expect(flattenMarkup(renderRipple(true, "#fff", 80, 3, "fixed"))).toContain("width:80px");
```

Both directions, so it fails if the argument is dropped *and* if the mode is ignored. Checked it actually bites by removing `scale` at the `renderRipple` call site in `_renderItem` — red, as intended.

src/floorplan-card.ts
superflyer11: You're right, and the comment is rewritten in cdcd1dd to make your argument instead of mine. `--fp-u` is unregistered, so it's substituted as a token stream and the `cqw` resolves at whichever descendant of `.plan` uses it — declaring it on `.plan` would have worked identically today. The reason to keep it on `.items` is that it stays correct if `--fp-u` is ever registered via `@property`, which would resolve the `cqw` at the declaring element and hand it the stage.

The `.plan`-not-`.stage` half of the comment is unchanged, since that one is load-bearing. Kept the note that adding `inline-size` containment to `.plan` doesn't disturb its own `100cqh` width, as that's the part that reads alarming.

src/editor-forms.ts
superflyer11: Renamed to `projectDisplayForm` in cdcd1dd, across `editor-forms.ts`, `editor.ts` and the tests.

superflyer11:
## Addressed — cdcd1dd

All six items, plus a correction to the blocking one's repro. Details are in the threads; this is the shape of it.

### The blocking fix

The room name's `font-size` is emitted inline only when it says something the stylesheet cannot — `overlayScale: plan`, or an explicit `labelSize`. `.area-label { font-size: 14px }` is back as a normal rule, so a default card is untouched. The decision lives in `areaLabelFontSize` beside the other style-sink helpers, which makes it unit-testable where `_renderAreaLabel` isn't.

**Your repro doesn't reproduce, though the finding stands.** I built the branch point, the reviewed commit and the fix, and injected into the card's shadow root the way card-mod does, on a default card:

| card-mod rule | branch point | reviewed | now |
| --- | --- | --- | --- |
| `.area-label { font-size: 28px }` | 14px | 14px | 14px |
| `.area-label { font-size: 28px !important }` | 28px | 28px | 28px |
| `ha-card .area-label { font-size: 28px }` | **28px** | **14px** | **28px** |

Row 3 is the regression, and it's fixed. Row 1 — the exact rule quoted in the review — was already losing before this branch: lit puts `static styles` in `adoptedStyleSheets`, which the cascade orders *after* a `<style>` injected into the same shadow root, so an equal-specificity rule loses on tree order regardless. Pre-existing; I've left it alone rather than quietly widen this branch. The people affected are the ones who out-specified the card's rule, which is what card-mod users write in practice.

### The README

Option 2, as you'd have taken. It carries your table and says the mode suits a card down to roughly two-thirds of its canvas width, past which the sizes have to come up — stated as a trade, not a free win. No floor: a `max(8px, …)` would stop the mode being proportional exactly where someone reached for it to be proportional, and would do it silently. Also documented `.area-label` in the hooks table and the new precedence rule on `Area.labelSize`.

### The follow-ups

Three were cheap enough to fold in rather than file: the `var(--fp-u, 1px)` fallback, the `renderRipple` threading assertion (checked it goes red when `scale` is dropped at a call site), and `projectRotationForm` → `projectDisplayForm`.

The `px` suffix is left as you suggested — it needs config plumbing into `areaForm` and six sliders already carry it, so it's a separate issue rather than this branch's job. Happy to open it.

I also rewrote the `--fp-u` placement comment: you were right that container-unit resolution isn't the reason, and it now argues from `@property` registration instead.

### Verified

Re-measured in a browser, not just the suite. At a card shape that letterboxes `.plan` to 420px inside a 900px stage, and at the canvas ratio, a room name measures **10.27 %** of the plan's wall-to-wall width both times — so the container choice still holds after the change. Under `plan`, the area name, item label and text all compute to exactly `N × planW/980`. The `var()` fallback is inert inside `.items`, to four decimals.

750 tests pass (744 before), typecheck clean.

nicosandller (approved):
## Approving

Re-checked on `cdcd1dd`. The blocking item is fixed, and you were right to correct my reproduction — I reran the three-build comparison myself rather than take the commit message's word, and it reproduces exactly:

| card-mod rule | pre-PR (`512fb47`) | reviewed (`3799998`) | now (`cdcd1dd`) |
| --- | --- | --- | --- |
| `.area-label { 28px }` | 14px | 14px | 14px |
| `.area-label { 28px !important }` | 28px | 28px | 28px |
| `ha-card .area-label { 28px }` | **28px** | **14px** | **28px** |

So my review asserted the regression using row 1, which never worked in any build — lit's `adoptedStyleSheets` are ordered after an injected `<style>`, so an equal-specificity rule loses on tree order regardless. The regression was real but lived in row 3, and I'd compared against a baseline I assumed rather than measured. Thanks for running it properly instead of just applying what I asked for; the fix addresses the case that actually broke.

Verified on the new head:

- **The `areaLabelFontSize` matrix**, measured in a browser — inline emitted only when it has something to say:

| mode | `labelSize` | inline? | computed |
| --- | --- | --- | --- |
| `fixed` (default) | unset | no | 14px (stylesheet — card-mod hook intact) |
| `fixed` | 26 | yes | 26px |
| `plan` | unset | yes | 9.96px = 14 canvas units |
| `plan` | 26 | yes | 18.50px = 26 canvas units |

- **750 tests pass, typecheck clean.**
- `overlayLength` carries the `var(--fp-u, 1px)` fallback, and the `renderRipple` assertion anchors the threading — that one would genuinely have caught a dropped argument.
- README now measures the trade instead of selling the mode, and names the two-thirds-of-canvas-width rule of thumb. Good call keeping proportionality rather than adding a floor; the docs make it a choice the reader can make.
- `projectDisplayForm` matches what it owns.
- The `--fp-u` comment now argues from `@property` registration, which is the right reason.

Deliberately left: the editor slider's `px` suffix. Agreed it's a follow-up — six sliders carry it and it needs config plumbing into `areaForm`.

One thing I chased and am **not** raising as an issue, so it doesn't surprise you later: under `plan`, a card whose container measures zero renders labels at `font-size: 0` rather than falling back. I hit it on a collapsed dev-harness pane. The whole card is 0×0 at that point so nothing is visible either way — noting it only so it's a known property rather than a future mystery bug report.

The part I most expected to be wrong — `.plan` vs `.stage` as the query container — was right the first time and stayed right through the refactor. Nice work.

PR #151, opened 2026-08-10
Sliders: a sensor per panel, and a biparting door that stacks instead of pocketing (closes #145)

### LINKED ISSUE #145: Two opening sensors for a biparting sliding door
## Feature request: Two opening sensors for a biparting sliding door

Hi,

First of all, thank you for Easy Floorplan. I really like it and have recently started using it.

I would like to suggest an improvement for `sliderStyle: biparting`.

For a real two-panel sliding door, I have **two independent door opening sensors** in Home Assistant, one for each panel:

```yaml
binary_sensor.sliding_door_left
binary_sensor.sliding_door_right
```

Currently, a sliding opening seems to support only one `entity`, which means that both panels have to share the same state.

Would it be possible to support two entities for a biparting sliding door, for example:

```yaml
motion: slide
sliderStyle: biparting
entity: binary_sensor.sliding_door_left
secondaryEntity: binary_sensor.sliding_door_right
```

The desired behaviour would simply be:

* Left sensor open → left panel is shown open.
* Right sensor open → right panel is shown open.
* Both open → both panels are shown open.
* Both closed → both panels are shown closed.

This would be particularly useful for **sliding patio doors / bay doors**, where each sliding panel has its own opening sensor.

Thank you for considering this feature!

superflyer11:
### The problem

@lepiaf56 has a two-panel patio slider with an independent contact on each leaf, and the card can say neither thing.

**One state for two panels.** A sliding opening carries one `entity`, and `biparting` draws both panels from it. Left open and right shut — the everyday case on a patio door you only ever open one side of — can only be drawn as both open or both shut.

**And it draws the wrong door.** `biparting` recesses each panel into the wall on its own side. That is a pocket door (*galandage*). The common patio slider has fixed panels at the jambs and the moving ones ride in front of them: nothing disappears into a wall, and even wide open the outer quarters stay glazed.

### The change

**`secondaryEntity`** binds the other panel — the field name an item already uses for its second reading:

```yaml
- id: bay
  type: window
  motion: slide
  sliderStyle: biparting-bypass
  entity: binary_sensor.sliding_door_left
  secondaryEntity: binary_sensor.sliding_door_right
```

Each panel then resolves its own amount, its own accent and its own outage. Unset, both panels stay on `entity` and every slider that exists today draws exactly as it did.

The plumbing is deliberately thin. `secondPanelOf()` hands the second panel to `resolveOpeningAmount` / `openingIsActive` as an opening in its own right, rather than threading a "which panel" argument down the resolver chain — so the second leaf inherits the rules the first one already had, including failing closed on `unavailable`. Per leaf, so one dropped sensor no longer shuts the whole opening. `invert` and the tap target stay shared: it is one opening.

**`sliderStyle: biparting-bypass`** draws the patio door. The opening splits into four — two fixed panels at the jambs on the front track, two moving ones on the back track — and the moving pair parts until each sits over the fixed panel on its side.

Travel is therefore a **quarter** of the opening rather than half, and the clear span tops out at the middle half however wide open the door is. That is the whole difference from `biparting`, and the reason it can't be had by clamping that one's travel: the outer quarters stay glazed, and on a plan that is the part you can't walk through.

**`sliderStyle: converging`** draws the same door without the fixed glass, after @lepiaf56 [clarified](https://github.com/nicosandller/easy-floorplan/pull/151#issuecomment-5251263471) that theirs has none: two leaves, both operable, overlapping each other rather than a fixed panel.

That is the **mirror image** of `biparting-bypass`, not a variant of it. The panels run *toward* each other and stack over the middle half, so what clears is a quarter at each jamb — where `biparting-bypass` keeps the outer quarters glazed and clears the middle. Neither can be had by clamping the other, so it joins it rather than replacing it; which one you want depends on whether your outer quarters are fixed glass.

Travel is a quarter here too, and that is the whole design of the style. A leaf can physically slide its own width, straight across its neighbour on the other track — but then two open leaves swap sides and cover the opening again, which is the one thing a plan must not draw. Parking each at the halfway point makes both-open read the way the door looks. The price: a single open leaf draws a quarter clear rather than the half it could reach. Under-promising the gap you can walk through beats inventing one that isn't there.

`secondaryEntity` needed nothing new for it — the second panel goes through `secondPanelOf` exactly as the biparting styles do. The style *list* was the fragile part: it was being retyped by hand in the editor's `toPatch`, so that check now asks `sliderStyleHasTwoPanels`, and `SliderStyle` is a named type rather than an inline union in three places.

Two smaller decisions, both commented:

- **The fixed panels are never accented, even wide open.** `bypass` does accent its fixed panel today and this deliberately doesn't follow it — the accent marks what has *moved*, and lighting these would accent exactly the half that is still shut. I left `bypass` alone rather than change a card that already exists.
- **Every two-panel style hides the slide-direction field.** A door that moves both leaves at once has no direction; `flipH` only swaps which panel each sensor drives. Switching a slider to a one-panel style drops `secondaryEntity`, the same way leaving `slide` already drops `sliderStyle`.

`biparting` is untouched and stays the default for anyone who really does have a pocket door, so no existing card changes on upgrade.

### Editor

**Style** gains *Biparting (over fixed panels)* and *Converging (both panels stack in the middle)* beside a relabelled *Biparting (into the walls)*. **Second panel** appears only on a two-panel slider that already has its first entity bound — a slider whose *second* panel alone had a sensor would be more confusing than useful — and takes the same contact/cover picker as the first.

### Verified

Unit tests can't see this, so it was measured on the built card in a browser, reading laid-out geometry back in canvas units. Opening spans **200–400**:

| | panel one | panel two | what clears |
| --- | --- | --- | --- |
| `converging`, both shut | 200–300 | 300–400 | nothing — it fills the opening |
| `converging`, both open | 250–350 | 250–350 | 200–250 and 350–400 |
| `converging`, left contact only | 250–350 | 300–400 | 200–250 |
| `biparting-bypass`, both open | 200–250 | 350–400 | 250–350 |
| `biparting`, both open | 100–200 | 400–500 | the whole opening |

`converging`'s two leaves land exactly on each other over the middle half, and the two patio styles come out as exact mirrors — which is the claim worth checking, since it is why both exist. Nothing crosses a jamb except `biparting`, which is supposed to. This also confirms `translateX(Npx)` on the SVG group still resolves in user units at the new quarter travel. Computed fills confirm only the leaves that moved wear the accent, and that `converging` has no fixed rect to leave unaccented.

949 tests pass (+41 across the three two-panel styles: per-panel travel in both directions, clamping, per-panel accent, the second state being ignored by the one-panel styles, and the form's field visibility and patches). README carries a table of all five styles, since which one you want is now a real choice.

Rebased onto current `main` after the review, which had moved six commits — including the shutter work in #153 and per-wall thickness in #157, both of which touch openings. The geometry above was re-measured afterwards and is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,8 +183,16 @@ window, a `blind` → a slider, a `garage` or `shutter` → a roll-up); adjust a
   tracks the position live. Everything else uses the on/off behavior above.
 - **Motion** — **swing** (default), **slide**, or **roll** (a slatted curtain that thins
   onto its track). Sliding openings take a **Style**: *single* (one panel into the wall),
-  *bypass* (two panels on parallel tracks), or *biparting* (two panels parting at the
-  middle). **Slide** sets the direction.
+  *bypass* (two panels on parallel tracks), *biparting (into the walls)* (two panels
+  parting at the middle, each recessing into its own wall — a pocket door), or
+  *biparting (over fixed panels)* (the same two panels stacking over a fixed panel at
+  each jamb — the usual patio / bay slider, where nothing leaves the opening and the
+  middle half is as clear as it gets). **Slide** sets the direction; a biparting slider
+  parts both ways, so it has none.
+- **One sensor per panel** — a biparting slider with a contact on each leaf takes a
+  **Second panel** entity, and then each panel opens and accents on its own state: left
+  open and right shut draws exactly that. Leave it empty and both panels follow the first
+  entity, as they always have. **Invert** covers both, and a tap still acts on the first.
 - **Orientation** — **Hinge** (left / right) and **Opens** (this side / other side) face a
   swing door any of four ways; they're pure mirrors (`flipH` / `flipV`), so the animation
   follows.
@@ -203,6 +211,9 @@ window, a `blind` → a slider, a `garage` or `shutter` → a roll-up); adjust a
 openings:
   # sliding window, patio-door style, driven by a cover
   - { id: patio, type: window, motion: slide, sliderStyle: biparting, x: 640, y: 500, length: 160, angle: 0, entity: cover.patio_door }
+  # a two-panel patio slider with a contact on each leaf: the panels stack over
+  # the fixed side panels, and each one follows its own sensor
+  - { id: bay, type: window, motion: slide, sliderStyle: biparting-bypass, x: 300, y: 500, length: 200, angle: 0, entity: binary_sensor.sliding_door_left, secondaryEntity: binary_sensor.sliding_door_right }
   # a swing door hinged on the right, opening into the other room
   - { id: hall, type: door, x: 300, y: 100, length: 80, angle: 0, flipH: true, flipV: true }
 ```
@@ -378,16 +389,17 @@ distorted anyway.
 | `length`      | number                      | Length along the wall.                                 |
 | `angle`       | number                      | Rotation in degrees.                                   |
 | `entity`      | string                      | Contact `binary_sensor` / `cover` driving open/closed (or `current_position` for partial). |
+| `secondaryEntity` | string                  | Biparting sliders: a second contact / `cover` for the other panel, so each leaf moves on its own state. Unset = both follow `entity`. |
 | `invert`      | boolean                     | Flip the open/closed interpretation.                   |
 | `activeColor` | string                      | Leaf/arc color while actively open (default primary).  |
 | `flipH`       | boolean                     | Mirror left↔right. Swing door: hinge jamb. Sliding: slide direction. |
 | `flipV`       | boolean                     | Mirror across the wall so a swing opening faces the other room. |
-| `sliderStyle` | `single` \| `bypass` \| `biparting` | With `motion: slide`: one panel (default), two stacking, or two centre-parting. |
 | `showShutterIcon` | boolean                 | Draw that icon (default `true` whenever both are bound). Editor: **Shutter icon**. Turning it off changes nothing about the gestures — for a plan where every window has a shutter and the icons start to shout. |
 | `shutterIcon` | string                      | Override the icon's glyph. Left unset it follows the shutter entity, whose default comes in an open/closed pair; an override is one glyph for both states, and colour still reports the state. |
 | `tapTarget`   | `opening` \| `shutter`      | With both entities bound, which one a tap acts on (default `opening`); the other moves to press-and-hold. Editor: **Tap opens**. Pointing it at the shutter opens the shutter's dialog — it does not drive the motor; set `tap_action: toggle` for that. |
 | `tap_action`  | ActionConfig                | Standard Lovelace action, acting on whichever entity `tapTarget` leads with (or on `shutterEntity` when it is the only one bound). By default an open/close `cover` toggles and everything else opens more-info. An action's own `entity` picks which of the two it acts on. |
 | `hold_action` / `double_tap_action` | ActionConfig | With both entities bound, hold opens the shutter's more-info by default — a tap is never retargeted at the shutter motor. Double-tap does nothing unless configured. |
+| `sliderStyle` | `single` \| `bypass` \| `biparting` \| `biparting-bypass` | With `motion: slide`: one panel (default), two stacking, two centre-parting into the walls, or two centre-parting over a fixed panel at each jamb. |
 
 ### Item (device)
 
--- a/README.md
+++ b/README.md
@@ -493,6 +493,12 @@ with nothing to configure. Windows behave the same way, so an open one spills li
 outside. A cover reporting a partial position opens a proportional gap, and at night the
 clearing a lit room holds against the dark reaches through the same doorways its pool does.
 
+A two-panel slider counts **both** its leaves, so a door with a sensor on each opens a gap
+when either one does. How wide follows what the style actually draws: `biparting` sends its
+leaves into the walls and can clear the whole opening, while `biparting-bypass` and
+`converging` keep theirs inside the frame and so clear at most half of it however wide open
+they are.
+
 Pools are drawn above room fills but below furniture and walls, so light reads as cast onto
 the floor. Furniture under a lit lamp picks up about half the cast, enough to read as lit
 without turning into the color of the light. Pools never intercept clicks.
--- a/README.md
+++ b/README.md
@@ -182,14 +182,22 @@ window, a `blind` → a slider, a `garage` or `shutter` → a roll-up); adjust a
 - **Partial** — a `cover` reporting `current_position` (0–100) is drawn partly open and
   tracks the position live. Everything else uses the on/off behavior above.
 - **Motion** — **swing** (default), **slide**, or **roll** (a slatted curtain that thins
-  onto its track). Sliding openings take a **Style**: *single* (one panel into the wall),
-  *bypass* (two panels on parallel tracks), *biparting (into the walls)* (two panels
-  parting at the middle, each recessing into its own wall — a pocket door), or
-  *biparting (over fixed panels)* (the same two panels stacking over a fixed panel at
-  each jamb — the usual patio / bay slider, where nothing leaves the opening and the
-  middle half is as clear as it gets). **Slide** sets the direction; a biparting slider
-  parts both ways, so it has none.
-- **One sensor per panel** — a biparting slider with a contact on each leaf takes a
+  onto its track). Sliding openings take a **Style**, and which one you want comes down to
+  where the panels go and what is left clear:
+
+  | Style | Panels | Where they go | What clears |
+  | --- | --- | --- | --- |
+  | *single* | one moving | into the wall | the whole opening |
+  | *bypass* | one moving, one fixed | behind the fixed one | half |
+  | *biparting (into the walls)* | two moving | each recesses into its own wall — a pocket door | the whole opening |
+  | *biparting (over fixed panels)* | two moving, two fixed | out onto a fixed panel at each jamb | the middle half |
+  | *converging* | two moving | toward each other, stacking in the middle | a quarter at each jamb |
+
+  The last two are both patio sliders and they are mirror images: pick *biparting (over
+  fixed panels)* if the outer quarters of your door are fixed glass, and *converging* if
+  every leaf slides. **Slide** sets the direction; a style that moves both panels has
+  none.
+- **One sensor per panel** — a two-panel slider with a contact on each leaf takes a
   **Second panel** entity, and then each panel opens and accents on its own state: left
   open and right shut draws exactly that. Leave it empty and both panels follow the first
   entity, as they always have. **Invert** covers both, and a tap still acts on the first.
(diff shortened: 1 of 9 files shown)


lepiaf56:
Thank you for working on this!

Just one clarification regarding the physical configuration of my sliding door.

There are **only two sliding panels** in the frame. There are no fixed side panels at either jamb.

When the door opens, the two panels slide in opposite directions and **stack directly on top of each other**. So the overlap is between the two moving panels themselves, not between a moving panel and a fixed side panel.

In other words, the configuration is:

```text
CLOSED
┌─────────────────┐
│       │         │
│       │         │
└─────────────────┘
        ↑
    two sliding panels

OPEN
┌─────────────────┐
│    ┌────────┐   │
│    │ panel  │   │
│    │   +    │   │
│    │ panel  │   │
│    └────────┘   │
└─────────────────┘
```

So I think my case is slightly different from the `biparting-bypass` configuration you describe, since there are no fixed panels.

The two independent opening sensors are still exactly what I need; I just wanted to clarify the physical geometry of the door.

superflyer11:
Thank you — that's a useful correction, and your diagram is unambiguous about the part that matters. You're right that it isn't the `biparting-bypass` configuration: I built that from your earlier note about the panels sliding in front of the fixed side panels, and with no fixed panels the geometry comes out as the mirror image rather than a variant.

`sliderStyle: converging` is now on the branch for exactly your door — two leaves, both operable, overlapping each other:

```yaml
- id: bay
  type: window
  motion: slide
  sliderStyle: converging
  entity: binary_sensor.sliding_door_left
  secondaryEntity: binary_sensor.sliding_door_right
```

Measured on the built card, for an opening spanning 200–400 in plan units:

| | panel one | panel two | what clears |
| --- | --- | --- | --- |
| both shut | 200–300 | 300–400 | nothing — it fills the opening |
| both open | 250–350 | 250–350 | 200–250 and 350–400 |
| left contact only | 250–350 | 300–400 | 200–250 |

Which is your second diagram: the two panels stacked together in the middle, a gap at each side.

`biparting-bypass` stays for the door with fixed glass at the jambs — it clears the middle instead of the ends, so the two really are opposites and neither can be had by clamping the other's travel. The **Style** dropdown now names all five, and the README has a table of where the panels go in each.

One thing worth confirming, since it's the only judgment call I had to make. Each leaf travels a **quarter** of the opening, so both-open parks them stacked in the centre as you drew. A leaf could physically slide its full width straight past its neighbour — but then two open leaves swap sides and cover the opening again, which would draw a shut door for a door that's wide open. The trade is that opening only the left leaf shows a quarter-width gap rather than the half you could actually walk through. If your door in practice slides one leaf right across the other, say so and I'll reconsider — but I'd expect the stacked-in-the-middle reading to be the one that matches what you see.

lepiaf56:
Yes, your quarter-opening approach makes sense for my door.

In theory, each leaf could travel up to about 1.50 m, which would allow one leaf to fully overlap the other.

However, in practice, I normally open each leaf by about 0.75 m. This gives me two simultaneous access openings, one on each side, which is actually the most useful configuration.

Also, the opening sensors only provide a binary state (open/closed); they do not report an intermediate position. So there is no need to reproduce the exact physical travel of each leaf.

For these reasons, having both leaves converge and stack in the centre, leaving an opening of approximately one quarter of the total width on each side, is a very coherent representation of my real-world use.

So I think the current `converging` behaviour is exactly what I need. 

You are doing remarkable work; allow me to express my sincere thanks.

nicosandller (changes requested):
Reviewed against #145. The design is right and the geometry is exactly as claimed — I re-measured it independently rather than take the table on trust, and every figure matches. **One blocking bug**, though, and it defeats the feature's headline case.

## Blocking: the second panel's sensor is never watched

`collectWatchedEntities` (`render.ts:95`) adds `secondaryEntity` for **items** but the openings loop only takes `entity` and `shutterEntity`:

```ts
for (const o of f.openings) {
  if (o.entity) ids.add(o.entity);
  if (o.shutterEntity) ids.add(o.shutterEntity);
  // o.secondaryEntity is missing
}
```

So when only the second panel's contact changes, `shouldUpdate` skips the render and the panel stays as it last drew. Measured on this branch, with the left sensor's state object carried over by identity the way HA delivers it:

```
WATCHED: ["binary_sensor.left"]
CARD RE-RENDERS WHEN THE RIGHT PANEL OPENS: false
```

That is exactly `Right sensor open → right panel is shown open` from the issue, not working. It looks intermittent rather than broken, because any *other* watched entity moving drags a render along with it and the panel catches up — which is the worst version of this bug to debug. The editor gates on the same set (`editor.ts:396`), so its preview freezes too.

The irony is that the comment three lines above this loop names the trap by number:

> Miss this and the plan is lit once and then frozen … the same trap entity-bound furniture (#82) and areas (#6) each fell into.

Fix is one line, plus a test — `collectWatchedEntities` is pure, so it pins directly:

```ts
if (o.secondaryEntity) ids.add(o.secondaryEntity);
```

Worth asserting `hassRenderInputsChanged` goes `true` when *only* the second sensor's object is replaced, since that is the actual failure and a set-membership check alone wouldn't have caught the "carried over by identity" subtlety.

## Everything else checks out

**Geometry, verified independently** — flattened `renderOpening` and read the panel rects back in canvas units for an opening spanning 200–400. Every number matches your table:

| | panel one | panel two |
|---|---|---|
| `converging`, both shut | 200–300 | 300–400 |
| `converging`, both open | 250–350 | 250–350 |
| `converging`, left only | 250–350 | 300–400 |
| `biparting-bypass`, both open | 200–250 | 350–400 |
| `biparting`, both open | 100–200 | 400–500 |

The mirror-image claim holds, and it's the thing worth having checked since it's the reason both styles exist: `biparting-bypass` clears 250–350 and keeps the ends glazed; `converging` clears 200–250 and 350–400 and stacks in the middle. Neither is the other with clamped travel.

**The `converging` travel decision is right.** Parking each leaf at the halfway point instead of letting it slide its full width means a single open leaf under-draws its gap — but the alternative has two open leaves swapping sides and re-covering the opening, which is worse on a plan than a conservative gap. Good call, and good that it's commented at the site rather than only in the PR body.

**`sliderStyleHasTwoPanels` in `toPatch`** is the right fix for the hand-retyped style list — a sixth style can't now be forgotten in the editor.

**Against the issue:** the reporter's literal YAML works (`biparting` is in the two-panel set, so `secondaryEntity` drives it), and their follow-up about non-pocket panels is answered twice over — `biparting-bypass` for fixed outer glass, `converging` for two operable leaves, after you checked with them which they had. `biparting` untouched and still the default, so nothing changes on upgrade.

`tsc` clean, 775 tests pass, build clean — all as stated.

## Non-blocking, for later

Now that #146 has merged, light passes through an opening in proportion to `resolveOpeningAmount(o, …)`, which reads `o.entity` only. A two-sensor door with just its *second* panel open will therefore draw that panel open while still blocking light. Same shape of gap as this one, in my code rather than yours — happy to follow up separately, just flagging it so it isn't discovered as a mystery later.

superflyer11:
Good catch, and thank you for measuring it rather than reading it — the entity would have been the one thing unit tests couldn't see me forget.

Fixed in `3728e73`, exactly as you suggested:

```ts
if (o.secondaryEntity) ids.add(o.secondaryEntity);
```

Reproduced your result first, then confirmed the fix on built cards, pushing `hass` the way HA does — a fresh state object for the sensor that moved, everything else carried across by identity:

| | watched entities | second leaf | re-rendered |
| --- | --- | --- | --- |
| before | `["binary_sensor.left"]` | 0 | **false** |
| after | `["binary_sensor.left","binary_sensor.right"]` | 300..400 → 250..350 | **true** |

That's the converging leaf travelling its full quarter off its own sensor, first leaf untouched.

Four tests, and I checked they bite — reverting the one line fails two of them. Your point about set membership not being enough was the useful half: `hassRenderInputsChanged` compares state objects by identity, so the assertion that matters is that it goes `true` when only the second sensor's object is replaced. Also covered: a one-sensor opening adds nothing, and a tick where neither leaf moved is still skipped.

### Rebased onto current main

`main` had moved six commits, including the shutter/gesture work in #153 and per-wall thickness in #157, both of which touch openings. Three conflicts, all "both sides added something":

- `OpeningStyle.shutter` grew `accent` / `flip` in #153 while this branch added `second` — kept both.
- The card's import list: #153 replaced `openingClickAction` with `openingActionForGesture` / `openingIsPressable`, so I took main's pair and added only `openingHasTwoPanels` / `secondPanelOf` rather than resurrecting the import main had dropped.
- The README opening table, where #153 added five gesture rows and this branch rewrote the `sliderStyle` row — kept main's rows, took this branch's row.

Re-measured the geometry afterwards rather than assuming, since #153 and #157 both land in this code:

```
converging — both shut          200..300   300..400
converging — both open          250..350   250..350
converging — left only          250..350   300..400
biparting-bypass — both open    200..250   350..400
biparting — both open           100..200   400..500
```

Identical to the table you verified, so nothing about the merge disturbed the drawing.

949 tests pass, typecheck and build clean.

### On the light gap

Left alone, as you scoped it — happy to take it if you'd rather not, just say. For whoever picks it up: `resolveOpeningAmount(o, …)` reading `o.entity` only means a door with just its *second* leaf open draws open while still blocking light. `secondPanelOf(o)` already exists for exactly this shape of problem, so it's likely `Math.max` of the two panels' amounts at the transmission site rather than anything structural.

nicosandller:
Pushed `3e26ffd` onto this branch — the light follow-up I flagged in review, so it lands with the feature rather than trailing it. Also confirming your `3728e73` fixes the blocking one: the second sensor is watched and I re-ran it, 954 tests green.

Two things were wrong, both from the light asking `entity` and stopping there.

**The second leaf was invisible to light.** A door whose only open panel was the second one drew that panel open and still blocked the pool outright — this PR's headline case, working everywhere except the light.

**Travel is not clearance.** `biparting` sends each leaf into a wall, so two open leaves clear the whole opening. Your two patio styles keep theirs inside the frame at a quarter of travel each, so even wide open they clear only half. Reading `amount` alone had `converging` passing twice the light it draws — a bug that only exists because the new styles exist, which is why it belongs here.

`openingClearFraction` owns both, and the numbers come from the geometry you measured:

| style | both open | one open |
| --- | --- | --- |
| `biparting` | all of it | half |
| `biparting-bypass` | half | a quarter |
| `converging` | half | a quarter |

Everything else keeps `amount` untouched — swing doors, roll-ups, `single`, `bypass`, and a single-sensor `biparting`, where both leaves share one amount and the mean of it is itself. No existing plan changes.

End to end, lamp above a wall with a converging door in it:

| | before | after |
| --- | --- | --- |
| both shut | dark | dark |
| left open | lit | lit |
| **right open only** | **dark** | **lit** |
| both open | lit | lit |

The shut row matters as much as the fixed one: cutting gaps must not decay into "the clipping is gone".

**One approximation I named rather than left implicit.** The gap is still cut *centred*, so for `converging` — whose leaves stack in the middle — it comes out the right size in the wrong place: the light through the wall is correct, where it lands is approximate. The soft radial falloff hides most of it. Fixing it properly means letting one opening contribute several intervals instead of one, which is a change to `wallsLightPassesThrough`'s contract rather than its arithmetic, and I did not want to reshape that inside your PR. Happy to do it separately if it ever shows.

Nothing else needed from you — this was mine to clean up after #146.

nicosandller (approved):

Issue #74, opened 2026-07-16
both window internal with external shutter


sononicola:
Can easy-floorplan handle something like this? I can't find it in the read.me example.
There is the internal window and then the external shutter to protect from sun.  

<img width="1379" height="948" alt="Image" src="https://github.com/user-attachments/assets/7da972c4-3eac-419e-b80f-fac5b6c86565" />

sononicola:
I'm sorry but this isn't what i asked, or at least I'm not able to use it, since i haven't any option in the below screen

<img width="730" height="702" alt="Image" src="https://github.com/user-attachments/assets/1c5bbb54-2bde-4c8a-906e-193cf6a5e877" />

Glad you added roller shutter for who asked in other issues, but I meant to add a simple external shutter like  in the photo above (that maybe claude can't see?) that has to work with a door binary sensor, not the cover entity of a roller shutter.

PR #153, opened 2026-08-10
feat(opening): reach the shutter behind the window, and show that it is there

Gifford47:
### Problem

An opening can bind two entities: `entity` for the sash and `shutterEntity` for the external roller shutter (#74). Both are drawn (`shutterAmount` / `shutterActive` / `shutterStyleOf`) and both are watched in `collectWatchedEntities`, but only `entity` was ever read for a press. Three things followed.

**The shutter was unreachable.** With both bound, the tap always targeted `entity`. Where that is a read-only contact `binary_sensor`, `openingClickAction()` returns `more-info` and the contact's dialog opens — the shutter drawn right there on the plan could be neither inspected nor moved.

**Shutter-only openings were inert.** An opening carrying only a `shutterEntity` never reacted at all: the `!o.entity` guard returned before anything could happen.

**Nothing said there were two entities.** Even once the shutter is reachable, the symbol still looks like a single thing, so whichever gesture reaches the second entity is something you would have to already know about to find.

### Change

**One resolver for every gesture.** `openingActionForGesture(o, gesture, featuresOf)` returns the action *and* the entity it acts on:

| gesture | default |
| --- | --- |
| tap | the opening — `toggle` for an open/close `cover`, else `more-info` |
| hold | `more-info` on the shutter — whichever entity the tap did not take |
| double tap | nothing |

A configured `tap_action` / `hold_action` / `double_tap_action` always wins, and its own `entity` field picks which of the two it acts on. `openingIsPressable` mirrors `itemIsInteractive` from #134 and gates the hit target, so a **shutter-only opening becomes pressable** while an opening with nothing bound stops claiming to be a button. Hold and double-tap timers are armed only when those gestures actually resolve, so an ordinary contact sensor still answers a tap immediately.

**Why the tap does not simply default to the shutter.** That was this PR's first proposal, and it is the one thing #47 argues against: a tap is the gesture people make by accident, and a shutter is real hardware that takes seconds to travel. Retargeting every tap at the shutter motor by default reintroduces exactly the accidental-hardware-move that #47 took out of the device defaults. The shutter is reached three other ways instead — all of them explicit:

- **press and hold** (the default once both entities are bound),
- **`tapTarget: shutter`** per opening, shown in the editor as **Tap opens**. Pointing the tap at the shutter opens its *dialog* rather than driving the motor; moving it is a further, separate decision that stays with `tap_action: toggle`,
- **the shutter's own icon**, below.

**The second entity is now visible.** An opening binding both draws the shutter entity's own icon beside it, and tapping that icon opens the shutter whatever the opening itself does. The glyph is resolved exactly as HA resolves it — registry override, then the icon on the state, then the domain/device-class default — and those defaults come in pairs, so a `shutter` cover reads `window-shutter-open` while open and `window-shutter` while shut. The accent repeats the same reading in colour. It is HTML in the overlay rather than SVG, for the two reasons the device badges are: it holds a real `ha-icon`, and it is sized in screen pixels, so it stays legible on a canvas of any scale.

**The shutter gets a reading and a look of its own.** `shutterInvert` (a reed contact on hinged shutters commonly reads `on` when the panels are *shut*, while the window behind it reads the other way round), `shutterActiveColor`, and `shutterFlipV` for hanging hinged panels on the sash's own side of the wall. Invert applies to the position *and* to the open-ish state, but the outage check comes first: an `unavailable` shutter must never invert into a wide-open one. Transit (`opening` / `closing`) stays active either way round.

**Editor.** Shutter invert / side / colour rows, the three gesture fields, and **Tap opens**. The gesture fields' defaults name what the card would *actually* do — the tap field reads `toggle` only for a cover that can open and close — so no field advertises a default the card does not take. The shutter badge is previewed on the canvas, inert there so it cannot swallow a selection click.

### The open question, answered

> Should long-press on an opening surface the other entity, the sash contact?

Yes — that is now the default whenever both entities are bound, and it works in both directions: hold always reaches whichever entity the tap did not.

### Notes

**Backwards compatible.** Every new key is optional, and a plan binding only `entity` renders and behaves exactly as before. Nothing needs to change in existing YAML.

**Now verified.** The earlier "could not run this locally" caveat no longer applies: `npx tsc --noEmit` is clean and `npx vitest run` passes **797 tests across 10 files**, 59 of them new — invert and sensor-outage behaviour, the full gesture-resolution table, the badge's position and icon resolution, the shutter side and accent in `renderOpening`, and the new form fields with patch round-tripping.

**About this branch's history.** The first two commits are superseded by the third: `_onOpeningClick` and its `_openMoreInfo` helper are gone, replaced by the gesture path above, and the README line they added is rewritten. The net diff against `main` is the final state.

**New config keys.** `tapTarget`, `shutterInvert`, `shutterActiveColor`, `shutterFlipV`, `tap_action`, `hold_action`, `double_tap_action` — all documented in the README's Opening table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,8 @@ distorted anyway.
 | `flipH`       | boolean                     | Mirror left↔right. Swing door: hinge jamb. Sliding: slide direction. |
 | `flipV`       | boolean                     | Mirror across the wall so a swing opening faces the other room. |
 | `sliderStyle` | `single` \| `bypass` \| `biparting` | With `motion: slide`: one panel (default), two stacking, or two centre-parting. |
+| `showShutterIcon` | boolean                 | Draw that icon (default `true` whenever both are bound). Editor: **Shutter icon**. Turning it off changes nothing about the gestures — for a plan where every window has a shutter and the icons start to shout. |
+| `shutterIcon` | string                      | Override the icon's glyph. Left unset it follows the shutter entity, whose default comes in an open/closed pair; an override is one glyph for both states, and colour still reports the state. |
 | `tapTarget`   | `opening` \| `shutter`      | With both entities bound, which one a tap acts on (default `opening`); the other moves to press-and-hold. Editor: **Tap opens**. Pointing it at the shutter opens the shutter's dialog — it does not drive the motor; set `tap_action: toggle` for that. |
 | `tap_action`  | ActionConfig                | Standard Lovelace action, acting on whichever entity `tapTarget` leads with (or on `shutterEntity` when it is the only one bound). By default an open/close `cover` toggles and everything else opens more-info. An action's own `entity` picks which of the two it acts on. |
 | `hold_action` / `double_tap_action` | ActionConfig | With both entities bound, hold opens the shutter's more-info by default — a tap is never retargeted at the shutter motor. Double-tap does nothing unless configured. |
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ distorted anyway.
 | `type`        | `door` \| `window`          | The kind of opening.                                   |
 | `motion`      | `swing` \| `slide` \| `roll` | How it moves: hinged (default), sliding panels, or a roll-up curtain (garage / roller shutter). |
 | `sash`        | `single` \| `double`        | Swing windows only: one full-width sash, or the classic two. Default `double`. |
-| `shutterEntity` | string                     | An external shutter over the same gap (`cover` or contact), with its own open/closed state. |
+| `shutterEntity` | string                     | An external shutter over the same gap (`cover` or contact), with its own open/closed state. Tapping the opening opens this entity's more-info dialog. |
 | `shutterStyle` | `swing` \| `roll`           | Louvered panels or a roll-up curtain. Defaults from the entity (contact → `swing`, `cover` → `roll`). |
 | `x`, `y`      | number                      | Center position.                                       |
 | `length`      | number                      | Length along the wall.                                 |
--- a/README.md
+++ b/README.md
@@ -368,8 +368,11 @@ distorted anyway.
 | `type`        | `door` \| `window`          | The kind of opening.                                   |
 | `motion`      | `swing` \| `slide` \| `roll` | How it moves: hinged (default), sliding panels, or a roll-up curtain (garage / roller shutter). |
 | `sash`        | `single` \| `double`        | Swing windows only: one full-width sash, or the classic two. Default `double`. |
-| `shutterEntity` | string                     | An external shutter over the same gap (`cover` or contact), with its own open/closed state. Tapping the opening opens this entity's more-info dialog. |
+| `shutterEntity` | string                     | An external shutter over the same gap (`cover` or contact), with its own open/closed state. With `entity` bound too, the card draws the shutter's own icon beside the opening — open/closed in both glyph and colour — and tapping that icon opens the shutter. |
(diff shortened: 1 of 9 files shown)


nicosandller (approved):
Approving. The design reasoning is sound and I checked the risky parts empirically rather than reading them — backwards compatibility, the outage/invert ordering, and the badge's rotation all hold up. Two non-blocking things below, one of which is a judgement call I'd rather you made than I did.

## Verified

**Backwards compatibility is clean.** Resolved the gesture table directly:

| opening | tap | hold |
| --- | --- | --- |
| `cover` only | `toggle` on the cover | — |
| contact only | `more-info` on the contact | — |
| both bound | `more-info` on the window | `more-info` on the shutter |
| both, `tapTarget: shutter` | `more-info` on the **shutter** | `more-info` on the window |
| shutter only | `toggle` on the shutter | — |
| nothing bound | not pressable | — |

Rows 1–2 are exactly what `main` does, and hold stays `undefined` wherever there is no second entity — so no hold timer is armed on an ordinary opening and nothing about an existing plan changes. Row 4 is the discipline worth calling out: even handed a cover with OPEN|CLOSE, the shutter-led tap resolves to `more-info`, not `toggle`. The #47 argument is actually implemented, not just described.

**Fail-closed ordering is right.** `shutterAmount` checks the outage before `invert` touches anything, so an `unavailable` shutter can't invert into a wide-open one. That's the specific trap and it's handled.

**The badge's rotation is correct** — I checked the push vector against `rotatePlanPoint` itself rather than by eye, and they agree exactly at every rotation:

```
rot   0   plan says {x:0,y:1}    badge pushes {x:0,y:1}
rot  90   plan says {x:-1,y:0}   badge pushes {x:-1,y:0}
rot 180   plan says {x:0,y:-1}   badge pushes {x:0,y:-1}
rot 270   plan says {x:1,y:0}    badge pushes {x:1,y:0}
```

`tsc` clean, build clean, **827** tests pass (the body says 797 — stale after the last two commits).

## Non-blocking

**1. `openingIsPressable` mirrors `itemIsInteractive`'s shape but not its substance.**

It asks `hasAction`, which only says a config exists and isn't `none`. `itemIsInteractive` deliberately does not do that — it asks each action for its own requirement, because a `toggle` with no entity, a `navigate` with no path and a `call-service` with no service all pass `hasAction` and then do nothing. Measured here:

```
opening with tap_action: {action: more-info} and NO entity bound
  → openingIsPressable: true
  → tap resolves to: {"config":{"action":"more-info"}}   ← no entity
```

So it takes `role="button"`, a tab stop and the press effect, and does nothing. Only reachable from hand-written YAML, so it's minor — but `role`/`tabindex` are *new* in this PR and this is the predicate gating them, which is what makes it worth closing rather than leaving. `gestureDoesSomething` in `actions.ts` is the existing shape to borrow.

**2. A shutter-only opening's tap drives the motor — your call.**

This one I'd rather flag than decide. The PR argues, correctly, that pointing a tap at a shutter should open its *dialog* and that moving it is a separate decision belonging to `tap_action: toggle`. `tapTarget: shutter` honours that exactly. But a shutter-only opening resolves its tap to `toggle`:

```
shutter-only tap: {"entity":"cover.sh","config":{"action":"toggle"}}
```

Both are "the tap acts on the shutter"; one refuses to move the hardware and the other moves it. It isn't a regression — `openingClickAction` has always toggled a capable `cover`, and that's pre-existing for openings — but these openings were previously **inert**, so this is a new path to moving hardware by accident, in the area #47 was about. Fine if that's what you want (the shutter *is* the whole opening there); worth being a decision rather than a side effect.

Neither blocks merge. Nice piece of work — the "reach the second entity by holding, in both directions" resolution is a better answer than the original question asked for.

Issue #147, opened 2026-08-09
Custom wall width just uses the smallest thickness, no matter the width value


NotTeflonDon:
This might already be possible, but I didn't see it when skimming the docs. I would like to be able to configure the width of the wall lines.

NotTeflonDon:
Nevermind, found the --fp-skin-wall-width style item. Thanks for all your hard work.

NotTeflonDon:
Although, if I use this:
```
card_mod:
  style: |
    ha-card {
      --fp-skin-wall-width: "5" !important;
    }
```
The number doesn't matter. It makes the walls super thin. Also, changing the theme removes the entire card-mod block.

superflyer11:
The quotes are doing it — `"5"` rather than `5`:

```yaml
card_mod:
  style: |
    ha-card {
      --fp-skin-wall-width: 5;
    }
```

`"5"` is a CSS *string*, which isn't a valid `stroke-width`. That makes the whole declaration invalid at computed-value time, and since `stroke-width` is an inherited property it then falls back to the initial `1` — so the wall draws at 1 unit instead of 8. That's the "super thin", and it's also why the number appeared not to matter: every quoted value fails in exactly the same way.

Worth knowing that the `8` in `var(--fp-skin-wall-width, 8)` can't rescue it either. A `var()` fallback only applies when the property is *unset*; a property set to something invalid is a different thing, and it's the failure you got.

Measured on a build, computed `stroke-width` on `.wall`:

| what you write | result |
| --- | --- |
| `--fp-skin-wall-width: 5` | 5px |
| `--fp-skin-wall-width: "5" !important` | 1px |

Also drop the `!important` — you don't need it on an unskinned card, and it's better to find out now if something else is competing.

One limit to keep in mind, since it isn't obvious: keep the value at **10 or below**. A doorway is a 12-unit hole punched through the wall layer, so a wall drawn wider than that isn't fully cleared by its own door and you get a sliver of wall across every opening. Only the stroke is skinnable — the geometry the mask and the door symbols are cut from is fixed.

---

[...]

Issue #122, opened 2026-08-05
Add "themes"


nicosandller:
I'm imagining community created themes for floorpans so we could add controls on the project AI to upload themes that can immediately apply to the floorplan and also have a few default themes to switch over.

By "upload" I dont mean upload a file somewhere in your HA, but maybe we can use a file selector and apply the content of the file directly onto our YAML ? not sure!

PR #156, opened 2026-08-11
Skins: put a palette above the card, not on it, so card-mod can still reach it (closes #155)

### LINKED ISSUE #155: Skins: a skin's inline style beats every documented card-mod token override
Every documented `--fp-skin-*` card-mod override is silently ignored once a plan sets `skin:` to anything other than `default`.

The README documents this as the way to restyle a plan without writing a skin:

```yaml
card_mod:
  style: |
    ha-card {
      --fp-skin-wall: #f2aa4c;
    }
```

That works on an unskinned card and does nothing at all on a skinned one.

### Why

[`floorplan-card.ts`](https://github.com/nicosandller/easy-floorplan/blob/main/src/floorplan-card.ts#L499) applies a skin as an inline `style` on `ha-card`:

```ts
<ha-card .header=${c.title ?? nothing} style=${skinStyle(c.skin) || nothing}>
```

card-mod's rule targets that same `ha-card`, and an inline style beats any non-`!important` rule — so the skin wins every time. The default skin is unaffected only because `skinStyle` returns `""` for it and no inline style is written at all, which is why this looks like it works until you pick a skin.

### Measured

Computed `stroke-width` on `.wall`, with the rule injected into the card's shadow root the way card-mod does it:

| card-mod rule | default skin | `skin: tron` |
| --- | --- | --- |
| *(none)* | 8px | 5px |
| `ha-card { --fp-skin-wall-width: 3 }` | **3px** ✓ | **5px** ✗ ignored |
| `ha-card { --fp-skin-wall-width: 3 !important }` | 3px ✓ | 3px ✓ |

`!important` is the only workaround today, and it's needed for every token individually.

This affects all 17 skin tokens on `odnetnin`, `pastel` and `tron` — not just the wall width. It came up in #147, where the reporter's own problem turned out to be a quoted value, but they'd have hit this next.

### Suggested fix

Move the palettes out of the inline style and into `static styles`, keyed off an attribute on the **host**:

```css
:host([data-skin="tron"]) { --fp-skin-wall: #7de3ff; /* … */ }
```

`:host` is strictly above `ha-card`, so a card-mod declaration on `ha-card` wins by inheritance — the same reason the default skin already behaves. The card sets `data-skin` only from a `findSkin` match, so an unknown `skin:` still can't put anything into the stylesheet. `skinTokens` already declares the defaults on `:host`, so this puts the palettes where the defaults already live.

### Out of scope / known limits

- A card-mod rule written against `:host` rather than `ha-card` still loses, both before and after: lit puts `static styles` in `adoptedStyleSheets`, which the cascade orders after an injected `<style>`, so an equal-specificity rule loses on tree order. `ha-card` is the documented idiom and it's the one worth making work.
- The editor applies `skinStyle` inline on its canvas ([`editor.ts`](https://github.com/nicosandller/easy-floorplan/blob/main/src/editor.ts#L2780)) too, but nothing card-mods the editor, so that one is fine as it is.

superflyer11:
### The problem

The README documents card-mod on `ha-card` as the way to restyle a plan without writing a skin:

```yaml
card_mod:
  style: |
    ha-card {
      --fp-skin-wall: #f2aa4c;
    }
```

That works on an unskinned card and does nothing at all on a skinned one — because a skin was applied as an inline `style` on that same `ha-card`, and an inline style beats any non-`!important` rule. Every one of the seventeen tokens, on `odnetnin`, `pastel` and `tron`. The default skin escaped only because `skinStyle` returns `""` for it, so nothing inline was written — which is precisely why the bug reads as "card-mod works" right up until someone picks a skin.

It surfaced in #147, where the reporter's own problem turned out to be a quoted value (`"5"` is a CSS string, not a length, so `stroke-width` went invalid and fell back to a hairline `1`). They'd have hit this next.

### The change

Declare the palettes on `:host` instead, keyed off a `data-skin` attribute the card sets from a `findSkin` match:

```css
:host([data-skin="tron"]) { --fp-skin-wall: #7de3ff; /* … */ }
```

`:host` is strictly above `ha-card`, so a card-mod declaration on that element wins by **inheritance** rather than having to win the cascade — the same reason overrides already worked on an unskinned plan. It also puts a palette where `skinTokens` already puts the defaults, so the two live together instead of at opposite ends of the render.

Three things worth stating, all commented:

- **The attribute is only ever a `findSkin` match**, so a hand-written `skin: nintendo` still puts nothing on the element and renders unskinned, exactly as before. The selector interpolates the id, so a test ties the palette to the registry's existing id-shape rule — a future skin called `a"],*` would otherwise be a stylesheet-wide rule.
- **`skinStyle` stays, for the editor.** Its canvas paints from an inline style it already builds, and nothing card-mods the editor. Noted at both ends so the two don't drift.
- **A card-mod rule against `:host` rather than `ha-card` still loses**, before and after. Lit puts `static styles` in `adoptedStyleSheets`, which the cascade orders *after* an injected `<style>`, so an equal-specificity rule loses on tree order. `ha-card` is the documented idiom and it's the one worth making work.

### Verified

Unit tests can't see this, so it was measured on built cards in a browser, injecting into the card's shadow root the way card-mod does. Computed `stroke-width` on `.wall`:

| card-mod rule | default | tron, before | tron, after |
| --- | --- | --- | --- |
| *(none)* | 8px | 5px | 5px |
| `ha-card { --fp-skin-wall-width: 3 }` | 3px | **5px — ignored** | **3px** |
| `ha-card { --fp-skin-wall-width: 3 !important }` | 3px | 3px | 3px |

The regression check mattered more than the fix, since this moves every skinned card's styling. Rendered output compared before and after across **unset / `default` / `odnetnin` / `pastel` / `tron` / an unknown id** — wall stroke, width and filter, plan paper, badge background and border:

```
identical across every skin
```

Including `tron`'s `drop-shadow` filter surviving the move, and the unknown id still falling back to unskinned.

743 tests pass (+5: the attribute's match rules, a host rule per non-default skin with its complete token set, the id-shape tie, and an assertion that nothing in the palettes selects `ha-card` — the whole fix in one line).

README gains the `data-skin` hook and the quoting trap from #147.

Closes #155.

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,7 +733,9 @@ so the canvas reads as the plan, and a **background image** still covers the ski
 ### Rolling your own
 
 A skin is a set of CSS custom properties, so [card-mod](#styling-hooks-card-mod) can set
-the same ones for the same result:
+the same ones for the same result — including on top of a skin, to change one thing about
+it rather than replace it. A skinned card also carries its id as `data-skin` on the card
+element, so a rule can apply to one skin only:
 
 ```yaml
 type: custom:easy-floorplan-card
@@ -746,6 +748,11 @@ card_mod:
       --fp-skin-accent: #f2aa4c;
     }
 ```
+
+Values are plain CSS, so quoting them breaks the declaration rather than setting it:
+`--fp-skin-wall-width: 5` works, `--fp-skin-wall-width: "5"` draws hairline walls. The
+`var()` default can't catch that — a fallback only applies to a property that is *unset*,
+never to one set to something invalid.
 
 | Token | Default | Paints |
 | --- | --- | --- |
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -70,7 +70,15 @@ import {
 } from "./render";
 import { deadSpacesCached } from "./dead-space";
 import type { Opening } from "./types";
-import { skinStyle, skinTokens, SKIN_ACCENT, SKIN_PAPER, SKIN_TEXT, SKIN_WALL } from "./skins";
+import {
+  skinAttribute,
+  skinPalettes,
+  skinTokens,
+  SKIN_ACCENT,
+  SKIN_PAPER,
+  SKIN_TEXT,
+  SKIN_WALL,
+} from "./skins";
 import { actionForGesture, executeAction, hasAction, itemIsInteractive } from "./actions";
 import { actionHandler } from "./action-handler";
 
@@ -154,6 +162,20 @@ export class FloorplanCard extends LitElement {
     const prev = changed.get("hass") as HomeAssistant | undefined;
     if (!prev || !this.hass) return true;
     return hassRenderInputsChanged(prev, this.hass, this._watchedEntities);
+  }
+
+  /**
+   * Carry the skin as an attribute on the host, where `skinPalettes` picks it
+   * up (issue #155). It has to be the host and not the template, because the
+   * point is to sit *above* the `<ha-card>` a card-mod rule targets — see
+   * skins.ts. Only ever a `findSkin` match, so an unrecognised `skin:` puts no
+   * attribute on the element at all.
+   */
+  protected willUpdate(changed: PropertyValues): void {
+    if (!changed.has("_config")) return;
+    const skin = skinAttribute(this._config?.skin);
+    if (skin) this.setAttribute("data-skin", skin);
+    else this.removeAttribute("data-skin");
   }
 
   public getCardSize(): number {
@@ -496,7 +518,9 @@ export class FloorplanCard extends LitElement {
            floor switcher and the card's own background follow it too — a Tron
            plan floating on a white card would read as a bug. Every token the
            card draws with is declared on :host, so this only ever overrides. -->
-      <ha-card .header=${c.title ?? nothing} style=${skinStyle(c.skin) || nothing}>
+      <!-- No skin style here: the palette comes from data-skin on the host, so
+           a card-mod rule on this element still wins (issue #155). -->
+      <ha-card .header=${c.title ?? nothing}>
         <div
           class="stage press-${pressEffectOf(c)}"
           style="aspect-ratio: ${dims.w} / ${dims.h};"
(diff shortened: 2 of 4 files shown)


nicosandller (approved):
## Approving

This does what #155 asks, and I checked that rather than taking the description's word for it — I built the branch and measured on real cards, since the failure this fixes is invisible to unit tests.

### Against the issue

#155 asks for the palettes to move off the inline `style` on `<ha-card>` and onto `:host([data-skin=…])`, so a card-mod rule on `ha-card` wins by inheritance. That is exactly what landed, including both edge cases the issue calls out: the attribute is only ever a `findSkin` match, and `skinStyle` stays for the editor.

Computed `stroke-width` on `.wall`, with the rule injected into the card's shadow root the way card-mod does:

| card-mod rule | unskinned | `default` | `tron` | `odnetnin` |
| *(none)* | 8px | 8px | 5px | 10px |
| `ha-card { --fp-skin-wall-width: 3 }` | 3px | 3px | **3px** | **3px** |
| …`!important` | 3px | 3px | 3px | 3px |

The two bold cells are the bug: they read 5px and 10px on `main` — the override silently ignored. `<ha-card>` now carries no inline style at all, and `data-skin` appears on the host only for a real non-default skin.

### Regression

This moves how every skinned card is styled, so that was the part worth checking hardest. Rendered output compared `main` vs this branch across unset / `default` / `odnetnin` / `pastel` / `tron` / an unknown id — wall stroke, width and filter, card paper, plan paper:

```
identical in every case, including tron's drop-shadow filter
and the unknown id still falling back to unskinned
```

One difference that is *not* a regression, in case it comes up: reading `--fp-skin-*` off the **host element** now returns the skin's values, where on `main` it returned the defaults. That is the whole point of the change — the tokens moved up one level — and everything that consumes them sits below `ha-card`, where the computed values are unchanged.

Also confirmed live skin switching still repaints: cycling skins on one card element gives the right walls each time, so the `keyed(c.skin, …)` rebuild from #122 still does its job with the attribute driving the palette instead of an inline style.

765 tests pass, typecheck and build clean, and it merges onto current `main` without conflicts.

### One nit, not worth blocking

The `static styles` comment says:

> Palettes come second: same specificity as the defaults, so tree order is what picks the skin.

Specificity is what picks it, not order — `:host([data-skin="x"])` is (0,2,0) against `:host`'s (0,1,0). I verified by adopting one stylesheet with the palette rule **first** and the bare `:host` default **second**; the palette still won.

Worth correcting because the comment implies reordering that array would break skinning, so the next person to touch it will preserve an ordering that does not matter — or file a bug when it does not. Same shape as the `--fp-u` placement note on #148: right call, wrong reason. Happy for it to ride along in a follow-up rather than another round here.

The README addition covering the quoting trap from #147 is a good catch to fold in — that is the failure the reporter actually hit.

PR #157, opened 2026-08-11
Walls: per-wall adjustable thickness

obbers:
## Summary
- Adds an optional `thickness` field to `Wall` (defaulting to the plan-wide `WALL_THICKNESS`) so individual walls can be drawn thicker or thinner than the rest of the plan.
- Adds a Thickness slider to the wall property panel in the editor.
- Applies in both the editor canvas and the live card.

## Test plan
- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm run build`
- [ ] Visual check in `npm run serve`: draw/select a wall, adjust its Thickness slider, confirm it renders thicker/thinner in both the editor canvas and the live-preview pane

### DIFF
--- a/src/editor-forms.ts
+++ b/src/editor-forms.ts
@@ -33,6 +33,7 @@ import {
 } from "./types";
 import {
   DEFAULT_LABEL_SIZE,
+  WALL_THICKNESS,
   badgeContentOf,
   domainIconAnimation,
   isPresenceEntity,
@@ -45,7 +46,7 @@ import {
   windowSash,
 } from "./render";
 import { defaultItemAction } from "./actions";
-import { DEFAULT_SKIN, SKINS, findSkin } from "./skins";
+import { DEFAULT_SKIN, SKINS, findSkin, MAX_SKIN_WALL_WIDTH } from "./skins";
 
 /** One ha-form schema item, extended with our label/helper (read by computeLabel). */
 export interface FormField {
@@ -861,9 +862,33 @@ export function wallForm(w: Wall): FormSpec {
     selector: { number: { mode: "box" } },
   });
   return {
-    fields: [coord("x1", "Start X"), coord("y1", "Start Y"), coord("x2", "End X"), coord("y2", "End Y")],
-    data: { x1: Math.round(w.x1), y1: Math.round(w.y1), x2: Math.round(w.x2), y2: Math.round(w.y2) },
-    toPatch: identity,
+    fields: [
+      coord("x1", "Start X"),
+      coord("y1", "Start Y"),
+      coord("x2", "End X"),
+      coord("y2", "End Y"),
+      {
+        name: "thickness",
+        label: "Thickness",
+        // Capped at MAX_SKIN_WALL_WIDTH, not a rounder number: past that a
+        // wall stops being fully cleared by its own door or window (the
+        // doorway mask's cut is sized off the shared WALL_THICKNESS
+        // constant, not per-wall — see render.ts's wallThickness).
+        selector: {
+          number: { min: 2, max: MAX_SKIN_WALL_WIDTH, step: 1, mode: "slider", unit_of_measurement: "px" },
+        },
+      },
+    ],
+    data: {
+      x1: Math.round(w.x1),
+      y1: Math.round(w.y1),
+      x2: Math.round(w.x2),
+      y2: Math.round(w.y2),
+      thickness: w.thickness ?? WALL_THICKNESS,
+    },
+    // Keep the default out of the YAML so untouched walls stay terse.
+    toPatch: (p) =>
+      "thickness" in p && p.thickness === WALL_THICKNESS ? { ...p, thickness: undefined } : p,
   };
 }
 
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -89,6 +89,7 @@ import {
   snapToWall,
   collectWatchedEntities,
   hassRenderInputsChanged,
+  wallStrokeStyle,
 } from "./render";
 import { deadSpacesCached } from "./dead-space";
 import { cssColor, cssColorOr, cssNumber, contrastText } from "./css-safe";
@@ -3316,7 +3317,7 @@ export class FloorplanCardEditor extends LitElement {
         <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
               class="wall ${selected ? "selected" : ""}"
               mask=${`url(#${this._wallMaskId})`}
-              stroke-width=${WALL_THICKNESS} stroke-linecap="round" />
+              style=${wallStrokeStyle(w.thickness)} stroke-linecap="round" />
         ${
           selected
             ? svg`
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -67,6 +67,7 @@ import {
   itemHiddenWhenInactive,
   itemLabelSize,
   areaLabelFontSize,
+  wallStrokeStyle,
   normalizeOverlayScale,
   overlayLength,
   hassRenderInputsChanged,
@@ -628,7 +629,7 @@ export class FloorplanCard extends LitElement {
                 (w) => svg`
                 <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
                       class="wall fp-wall" data-id=${cssIdent(w.id) ?? nothing}
-                      stroke-width=${WALL_THICKNESS} stroke-linecap="round" />`
+                      style=${wallStrokeStyle(w.thickness)} stroke-linecap="round" />`
               )}
             </g>
             <!-- Room outlines, above the walls they trace. An area polygon runs
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,6 +43,15 @@ export interface Wall {
   y1: number;
   x2: number;
   y2: number;
+  /**
+   * Stroke width in virtual units. Defaults to {@link WALL_THICKNESS}
+   * (render.ts) when unset, and clamped there (`wallThickness`) to at most
+   * `MAX_SKIN_WALL_WIDTH` (skins.ts): the doorway mask that cuts this wall's
+   * own doors and windows is sized off the shared `WALL_THICKNESS` constant,
+   * not per-wall, so a wall drawn wider than that ceiling would not be fully
+   * cleared by its own opening. Raise the ceiling only alongside that mask.
+   */
+  thickness?: number;
 }
 
 export type OpeningType = "door" | "window";
--- a/src/editor-forms.test.ts
+++ b/src/editor-forms.test.ts
@@ -22,7 +22,7 @@ import {
 import type { FormField } from "./editor-forms";
 import type { Area, Opening, FloorItem, Floor, FloorplanCardConfig } from "./types";
 import { DEFAULT_GLOW_RADIUS, DEFAULT_PRESS_EFFECT } from "./types";
-import { DEFAULT_SKIN, SKINS } from "./skins";
+import { DEFAULT_SKIN, SKINS, MAX_SKIN_WALL_WIDTH } from "./skins";
 
 const fields: FormField[] = [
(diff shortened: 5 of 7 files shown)


src/floorplan-card.ts
```diff
@@ -475,7 +475,7 @@ export class FloorplanCard extends LitElement {
                 (w) => svg`
                 <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
                       class="wall fp-wall" data-id=${cssIdent(w.id) ?? nothing}
-                      stroke-width=${WALL_THICKNESS} stroke-linecap="round" />`
+                      stroke-width=${w.thickness ?? WALL_THICKNESS} stroke-linecap="round" />`
```
nicosandller: **This has no effect once rebased on `main`.**

`.wall` now carries `stroke-width: var(--fp-skin-wall-width, 8)` (Skins, #122), and CSS beats a presentation attribute. Measured on the merged result: walls with `thickness` 3, unset and 24 all compute to `8px`.

The pattern #148 landed for `.area-label` fits here — inline only when it has something to say, so an untouched wall still follows the skin:

```ts
style=${w.thickness !== undefined ? `stroke-width:${wallThickness(w.thickness)}` : nothing}
```

src/editor.ts
```diff
@@ -3271,7 +3271,7 @@ export class FloorplanCardEditor extends LitElement {
         <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
               class="wall ${selected ? "selected" : ""}"
               mask=${`url(#${this._wallMaskId})`}
-              stroke-width=${WALL_THICKNESS} stroke-linecap="round" />
+              stroke-width=${w.thickness ?? WALL_THICKNESS} stroke-linecap="round" />
```
nicosandller: Same as the card: the editor's `.wall` rule also sets `stroke-width: var(--fp-skin-wall-width, 8)` on `main` (`src/editor.ts:4388`), so this attribute is overridden too. Both surfaces need the same fix, and they should keep using the same one so the editor keeps matching the card.

src/editor-forms.ts
```diff
@@ -799,9 +800,27 @@ export function wallForm(w: Wall): FormSpec {
...
-    toPatch: identity,
+    fields: [
+      coord("x1", "Start X"),
+      coord("y1", "Start Y"),
+      coord("x2", "End X"),
+      coord("y2", "End Y"),
+      {
+        name: "thickness",
+        label: "Thickness",
+        selector: { number: { min: 2, max: 30, step: 1, mode: "slider", unit_of_measurement: "px" } },
```
nicosandller: **A slider to 30 outruns the doorway mask.**

`renderWallMask` cuts each opening as `WALL_THICKNESS + 4` = 12 units, and that constant isn't per-wall. Measured with a 24-unit wall and a door on it: 6 units of wall remain across the doorway on each side.

`MAX_SKIN_WALL_WIDTH` in `src/skins.ts` caps skin wall width at 10 for exactly this reason, and explains it in the comment. Either thread the wall's thickness into the mask so each cut matches the wall it's on, or cap this at 10 for now and leave the mask work as a follow-up.

Separately: `unit_of_measurement: "px"` — these are canvas units. Six other sliders already say px, so I'd leave it, but it's inherited rather than intended.

src/types.ts
```diff
@@ -43,6 +43,8 @@ export interface Wall {
   y1: number;
   x2: number;
   y2: number;
+  /** Stroke width in virtual units. Defaults to {@link WALL_THICKNESS} (render.ts) when unset. */
+  thickness?: number;
```
nicosandller, line 54: Worth expanding this doc to say what the field can't do yet, since it isn't obvious from the type: above ~12 units the doorway mask (`WALL_THICKNESS + 4`) stops clearing the wall, so doors and windows leave slivers. Whatever ceiling you land on, stating it here is what stops the next person raising the slider's max without finding the mask.

The rest of the type addition is right — optional, defaulted at the read site, documented in virtual units.

nicosandller (changes requested):
## Requesting changes — the feature doesn't take effect on `main`

The shape of this is good: small, focused, `toPatch` correctly keeps the default out of the YAML, and the form test is the right one to write. But this branch is **8 commits behind `main`**, and one of those commits changes the outcome completely.

I built the merge of this branch with current `main` and measured it. The unchecked visual-check box in your test plan is, I think, exactly the right instinct.

### 1. `stroke-width` as a presentation attribute is overridden by CSS (blocking)

Skins (#122, `d22c96e`) added this to `.wall`, in both the card and the editor:

```css
.wall {
  stroke-width: var(--fp-skin-wall-width, 8);
}
```

CSS beats a presentation attribute, so setting the attribute has no effect. Measured on the merged result, three walls at `thickness: 3`, unset, and `24`:

| wall | attribute emitted | **computed** |
| --- | --- | --- |
| `thin` (3) | `3` | **8px** |
| `norm` (unset) | `8` | **8px** |
| `thick` (24) | `24` | **8px** |

Every wall renders at 8 regardless. It works on this branch's base only because `.wall` had no `stroke-width` rule back then — the comment `main` now carries on that rule says so directly: *"CSS beats the presentation attribute, so the skin sets the wall's weight."*

The fix mirrors a pattern the repo adopted last week in #148 for `.area-label`: write the value inline **only when it has something to say**, so an untouched wall keeps following the skin and an explicit thickness overrides it.

```ts
style=${w.thickness !== undefined ? `stroke-width:${wallThickness(w.thickness)}` : nothing}
```

Please rebase on `main` first — the merge is clean, so this is only a rebase, not a rework.

### 2. `WALL_THICKNESS` is shared geometry, not just a stroke width (blocking)

`renderWallMask` punches each doorway as a rect of `WALL_THICKNESS + 4` = **12 units**. That constant is not per-wall, so a wall drawn thicker than 12 is not fully cut by its own doors and windows. Measured on the merged result with the override forced past, a 24-unit wall with a door on it:

```
doorway cut:  height 12 units
wall stroke:  24 units
leftover:     6 units of wall across the doorway, on each side
```

The repo has already reasoned about this exact hazard — `MAX_SKIN_WALL_WIDTH` in `src/skins.ts` caps skin wall width at **10** and explains why:

> *"a wall wider than the hole would not be fully cleared by its own door, and every doorway would show a sliver of wall."*

This slider goes to 30. So per-wall thickness needs the mask to follow: `renderWallMask` should size each opening's cut from the wall that opening sits on, rather than from the global constant. Worth checking the opening symbols too — `renderOpening` takes its geometry from the same constant.

If you'd rather keep this PR small, capping the slider at 10 and clamping the same at the sink would ship something correct today, with the mask work as a follow-up. That's a legitimate choice — just not a slider to 30 with a 12-unit hole.

### Non-blocking

- **No clamp at the render sink.** House convention is explicit about this — see `itemLabelSize` and the test named *"clamp at the style sink"* from the #62 review, and `areaLabelSize` alongside it. The slider bounds 2–30 but YAML is unbounded, and `thickness: -5` or a string goes straight into the attribute. A `wallThickness(v)` helper in `render.ts` next to the others would match.
- **Hit area doesn't follow the wall.** `HIT.wall` is a fixed 11 and `.wall-hit` has a fixed stroke, so a 24-unit wall isn't clickable across its visible width and a 3-unit wall has an oversized grab area. Minor, but it's the kind of thing that reads as a bug once the widths are visibly different.
- **`unit_of_measurement: "px"`** — these are canvas units, not px. Six other sliders already carry the same wart so I wouldn't fix it here, but #148 just made the units question live in this codebase, so it's worth knowing it's inherited rather than intended.
- **Tests pass while the feature does nothing.** Not a criticism of the tests you wrote — they're correct for what they cover. It's the argument for the visual check being the one that mattered here.

---

Happy to re-review quickly once it's rebased. The underlying idea is a reasonable one and the diff is close to right; it's the base that moved under it.

obbers:
Rebased on main and addressed the review:

1. **stroke-width overridden by CSS (blocking)** — fixed. Added `wallStrokeStyle()`/`wallThickness()` in `render.ts`, mirroring the `areaLabelFontSize` pattern from #148: the width is written as an inline `style` only when a wall states its own thickness, so an untouched wall still follows the skin's `--fp-skin-wall-width` and an explicit thickness now actually overrides it. Applied identically in both the card and the editor.

2. **Doorway mask geometry (blocking)** — took the option to keep this small: capped the slider (and the render-sink clamp) at `MAX_SKIN_WALL_WIDTH` (10), the same ceiling skins already observe and for the same reason. A wall can no longer be drawn wider than the doorway mask's 12-unit cut clears, so no slivers. Left the per-wall mask work as a possible follow-up if a wider ceiling is ever wanted.

Non-blocking:
- Added the sink clamp (`wallThickness`) so a hand-edited/imported config that skips the slider can't push `thickness` out of range or into a style-injection payload — same guard `itemLabelSize`/`areaLabelSize` use.
- Left `unit_of_measurement: "px"` and the hit-area sizing alone per your notes — the thickness cap means the hit-area mismatch you measured (24-unit wall not fully clickable) no longer applies at the new ceiling, and the pre-existing generous tolerance for thin walls predates this PR.
- Expanded the `Wall.thickness` doc comment to state the ceiling and why.

Added test coverage for `wallThickness`/`wallStrokeStyle` (defaulting, clamping both directions, style-injection neutralization, and the "emits nothing when unset" case) and updated the two `wallForm` tests that used now-out-of-range values, plus a new test asserting the slider's `max` is `MAX_SKIN_WALL_WIDTH`.

`npm run typecheck`, `npm test` (760 passing), and `npm run build` all pass. I don't have a way to drive an interactive browser in my environment, so I can't tick the visual-check box myself this round either — but the specific failure you measured (computed 8px regardless of the attribute) is now covered directly by the `wallStrokeStyle` unit tests, and the mask hazard is closed by construction (thickness can no longer exceed what the 12-unit cut clears). Worth an eyeball pass on your end before merging.

nicosandller (approved):
## Approving

Both blocking items are resolved, and I re-measured rather than reading the diff — the override was the kind of bug that only shows up when you look at a computed style.

### The stroke-width override

`wallStrokeStyle` writes inline only when the wall asks for a width, so the skin keeps the untouched case. Measured across three skins, four walls:

| wall | default | `tron` (skin 5) | `odnetnin` (skin 10) |
| --- | --- | --- | --- |
| unset | 8px | **5px** | **10px** |
| `thickness: 2` | 2px | 2px | 2px |
| `thickness: 10` | 10px | 10px | 10px |
| `thickness: 30` | **10px** | **10px** | **10px** |

Row 1 is the one I'd have most expected a fix to get wrong — an untouched wall still follows the skin instead of being pinned to 8. That falls out of returning `""` rather than a default, which is the right shape and mirrors `areaLabelFontSize`. The editor produces the identical four values (`5 / 10 / 10 / 2`), so the two surfaces agree.

### The doorway mask

Capping at `MAX_SKIN_WALL_WIDTH` rather than a rounder number is the right call, and it makes the constant mean one thing across skins and per-wall widths. At the new maximum:

```
wall at max thickness: 10 units
doorway cut:           12 units
clearance:             1 unit each side — fully cleared
```

So the slider can no longer reach a width the mask can't clear. Tying the ceiling to that constant also means raising it later is one edit in one place, and `Wall.thickness`'s doc now says to raise it only alongside the mask — that note is what stops the next person moving the slider max without finding `renderWallMask`.

### The rest

- `wallThickness` clamps at the sink, next to `itemLabelSize` / `areaLabelSize`, with the injection case covered (`"5px;color:red"` → 8).
- Tests are the right ones: the floor, the cap, the injection payload, and — the one that matters most — that `wallStrokeStyle(undefined)` is `""`, which is what protects the skin.
- 760 tests pass; typecheck and build clean; no console errors.

Left as follow-ups, and I agree with both: the editor's hit area still doesn't track thickness, and the `px` suffix on the slider is inherited from six other sliders.

Good turnaround on this one — the fix went at the cause rather than at my description of it, and tying the ceiling to `MAX_SKIN_WALL_WIDTH` is a better answer than the arbitrary cap I suggested.

PR #158, opened 2026-08-11
Areas: tap a room to zoom in on it

obbers:
## Summary
- Tapping an area's fill on the live card zooms the plan in to that room; tapping it again (or a zoom-out button that appears top-left) returns to the full view. Switching floors resets the zoom.
- The zoom is one shared CSS transform around both the SVG and the HTML overlay, framed on the tapped area's bounding box (padded so the room isn't cropped edge-to-edge, capped at 4x, never zooming out past 1x) — computed with the same `rotatePlanPoint` math the overlay already uses so it stays correct on a rotated plan.
- Device badges, text, and area-name labels counter-scale against the zoom (via an inherited `--fp-inv-zoom` custom property) so they stay a constant, legible screen size instead of growing along with the zoomed room.

## Test plan
- [x] `npm run typecheck`
- [x] `npm test` (adds `areaZoomTransform` coverage: identity/empty polygon, scale cap, floor-at-1x for an oversized room, and rotation handling)
- [x] `npm run build`
- [x] Manually verified against a real Home Assistant instance: tapping a room zooms in, the zoom-out button returns to the full view, and device badges hold their size while the room scales

### DIFF
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -79,6 +79,8 @@ import {
   rotatedCanvasSize,
   rotatePlanPoint,
   planRotationTransform,
+  areaZoomTransform,
+  IDENTITY_ZOOM,
   type PlanRotation,
 } from "./render";
 import { deadSpacesCached } from "./dead-space";
@@ -113,6 +115,8 @@ export class FloorplanCard extends LitElement {
   @state() private _config?: FloorplanCardConfig;
   /** View-state: which floor is shown. Never persisted to config. */
   @state() private _activeFloorId?: string;
+  /** View-state: which area (if any) the plan is zoomed in to. Never persisted. */
+  @state() private _zoomedAreaId?: string;
   private readonly _wallMaskId = `fp-wall-mask-${FloorplanCard._nextWallMaskId++}`;
   /** Prefix for this card's glow gradient ids, unique per instance (issue #6). */
   private readonly _glowIdBase = `fp-glow-${FloorplanCard._nextGlowId++}`;
@@ -251,6 +255,11 @@ export class FloorplanCard extends LitElement {
         })
       );
     }
+  }
+
+  /** Tapping a room zooms the plan in to it; tapping the same room again zooms back out. */
+  private _onAreaClick(a: Area): void {
+    this._zoomedAreaId = this._zoomedAreaId === a.id ? undefined : a.id;
   }
 
   private _renderBadge(item: FloorItem, scale: OverlayScale): TemplateResult {
@@ -465,7 +474,7 @@ export class FloorplanCard extends LitElement {
         style="left:${(p.x / d.w) * 100}%; top:${(p.y / d.h) * 100}%;
                font-size:${overlayLength(cssNumber(t.size, DEFAULT_TEXT_SIZE), scale)};
                color:${cssColorOr(t.color, SKIN_TEXT)};
-               transform:translate(-50%,-50%) rotate(${cssNumber(t.angle, 0)}deg);"
+               transform:translate(-50%,-50%) scale(var(--fp-inv-zoom,1)) rotate(${cssNumber(t.angle, 0)}deg);"
       >
         ${t.text}
       </div>
@@ -530,6 +539,12 @@ export class FloorplanCard extends LitElement {
           lightWalls
         )
       : nothing;
+    // Zoom-to-room (tap an area). Both the SVG and the HTML overlay live
+    // inside one transformed wrapper below, so the two layers — positioned
+    // completely differently (a group transform vs. per-point left/top%) —
+    // reframe identically instead of drifting apart under zoom.
+    const zoomedArea = active.areas?.find((a) => a.id === this._zoomedAreaId);
+    const zoom = zoomedArea ? areaZoomTransform(zoomedArea.points, c.width, c.height, rot) : IDENTITY_ZOOM;
     return html`
       <!-- The skin (issue #122) rides on the card rather than on .plan, so the
            floor switcher and the card's own background follow it too — a Tron
@@ -570,6 +585,15 @@ export class FloorplanCard extends LitElement {
                count), "meet" letterboxes the SVG away from the overlay and
                every icon drifts off the wall it was placed on, while "none"
                stretches both layers identically: distorted, but aligned. -->
+          <!-- Zoom-to-room (tap an area). One wrapper around both the SVG and
+               the HTML overlay so a CSS transform here reframes both layers
+               identically — see areaZoomTransform. Wraps the keyed() skin
+               block below rather than sitting inside it, so a skin change
+               (which rebuilds that subtree) never disturbs this transform. -->
+          <div
+            class="plan-zoom"
+            style="transform: translate(${zoom.txPercent}%, ${zoom.tyPercent}%) scale(${zoom.scale});"
+          >
           <!-- Keyed on the skin (issue #122). A skin changes custom properties on
                an ancestor, and Chromium does not repaint an SVG element whose
                colour comes from a var() inside a presentation attribute or an
@@ -588,8 +612,11 @@ export class FloorplanCard extends LitElement {
                           preserveAspectRatio=${imageFitRatio(active.imageFit)}
                           opacity=${active.imageOpacity ?? 1} />`
               : nothing}
-            ${active.areas?.map((a) =>
-              renderArea(a, areaColor(a, a.entity ? this.hass?.states[a.entity]?.state : undefined))
+            ${active.areas?.map(
+              (a) =>
+                svg`<g class="area-tap-target" @click=${() => this._onAreaClick(a)}>
+                  ${renderArea(a, areaColor(a, a.entity ? this.hass?.states[a.entity]?.state : undefined))}
+                </g>`
             )}
             <!-- Dead spaces (issue #88): the regions the walls seal off that no
                  door or window reaches, hatched. Above the room fills, so a
@@ -722,7 +749,7 @@ export class FloorplanCard extends LitElement {
             </g>
           </svg>`
           )}
-          <div class="items">
+          <div class="items" style="--fp-inv-zoom:${1 / zoom.scale};">
             ${active.areas?.map((a) => this._renderAreaLabel(a, c, rot, scale))}
             ${active.texts.map((t) => this._renderText(t, c, rot, scale))}
             ${repeat(
@@ -743,6 +770,16 @@ export class FloorplanCard extends LitElement {
             )}
           </div>
           </div>
+          ${zoom.scale > 1
+            ? html`<button
+                class="zoom-out"
+                title="Zoom out"
+                aria-label="Zoom out"
+                @click=${() => (this._zoomedAreaId = undefined)}
+              >
+                <ha-icon icon="mdi:magnify-minus-outline"></ha-icon>
+              </button>`
+            : nothing}
           ${floors.length > 1 ? this._renderFloorSwitcher(floors, active) : nothing}
         </div>
       </ha-card>
@@ -766,6 +803,8 @@ export class FloorplanCard extends LitElement {
                 this._activeFloorId = f.id;
                 // Remember it for the next element the editor preview builds.
                 lastViewedFloor.set(floorMemoryKey(floors), f.id);
+                // A room on the floor just left is meaningless on the new one.
+                this._zoomedAreaId = undefined;
               }}
             >
               ${f.short || f.name}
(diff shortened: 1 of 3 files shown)


src/render.ts
```diff
@@ -2018,6 +2018,65 @@ export function polygonCentroid(points: readonly AreaPoint[]): { x: number; y: n
...
+  const maxY = Math.max(...ys);
+  const pad = Math.max(maxX - minX, maxY - minY) * padFrac;
+  const bw = Math.max(maxX - minX + pad * 2, 1);
+  const bh = Math.max(maxY - minY + pad * 2, 1);
+  const scale = Math.max(1, Math.min(maxScale, Math.min(d.w / bw, d.h / bh)));
+  const cxFrac = (minX + maxX) / 2 / d.w;
+  const cyFrac = (minY + maxY) / 2 / d.h;
+  return {
+    scale,
+    txPercent: 50 - scale * cxFrac * 100,
```
nicosandller: **The translation is never clamped to the plan's edges, so a zoomed room can pan the plan off the viewport.**

Centring the bbox is right, but nothing stops the result from moving the plan's own edge inside the visible box. Measured on this branch (canvas 980x700, card 698px wide, area `(40,40)-(140,110)`):

```
transform: translate(13.2653%, 7.14286%) scale(4)
blank left: 93px    blank top: 36px
```

That is bare card background where the plan should be. The usual clamp:

```ts
const clamp = (t: number) => Math.min(0, Math.max(100 * (1 - scale), t));
return {
  scale,
  txPercent: clamp(50 - scale * cxFrac * 100),
  tyPercent: clamp(50 - scale * cyFrac * 100),
};
```

A room near an edge then frames as close as it can without uncovering the plan — and it fixes the `scale === 1` case below for free, since the clamp range collapses to `[0, 0]`.

Measure with `transition: none` forced on `.plan-zoom`, or a throttled tab will hand you the pre-transition values.

src/render.ts
```diff
@@ -2018,6 +2018,65 @@ export function polygonCentroid(points: readonly AreaPoint[]): { x: number; y: n
...
+  const xs = rotated.map((p) => p.x);
+  const ys = rotated.map((p) => p.y);
+  const minX = Math.min(...xs);
+  const maxX = Math.max(...xs);
+  const minY = Math.min(...ys);
+  const maxY = Math.max(...ys);
+  const pad = Math.max(maxX - minX, maxY - minY) * padFrac;
+  const bw = Math.max(maxX - minX + pad * 2, 1);
+  const bh = Math.max(maxY - minY + pad * 2, 1);
+  const scale = Math.max(1, Math.min(maxScale, Math.min(d.w / bw, d.h / bh)));
```
nicosandller, line 2365: **When the scale floors to 1, the pan still applies — so tapping a tall room slides the plan for no benefit.**

`Math.max(1, …)` correctly refuses to zoom out, but the translate below is computed from the room centre regardless, so the plan moves without ever getting closer.

Measured (canvas 980x700, card 698px wide, area `(40,40)-(400,660)`): padded bbox is 546x806, and 806 > 700 so scale floors to 1 — but `txPercent` is still **27.551**, which shifts the plan 192px right, blanks 192px on the left, and pushes the right edge out of view. From a single tap, with no zoom to show for it.

The clamp suggested on the return statement fixes this too: at `scale === 1` the allowed range is `[0, 0]`, so tx is forced to 0.

src/floorplan-card.ts
```diff
@@ -567,7 +589,7 @@ export class FloorplanCard extends LitElement {
             }
             </g>
           </svg>
-          <div class="items">
+          <div class="items" style="--fp-inv-zoom:${1 / zoom.scale};">
```
nicosandller, line 752: **`1 / zoom.scale` is written raw — a NaN here takes out every device badge on the plan.**

One non-numeric coordinate in a hand-edited area (`{x: oops, y: 40}`) makes `Math.min`/`Math.max` return NaN, so `scale` is NaN and this emits `--fp-inv-zoom:NaN`. Measured on this branch after tapping such an area:

```
.plan-zoom transform : ""     (invalid, dropped)
--fp-inv-zoom        : NaN
.item computed       : none   <-- translate(-50%,-50%) lost too
```

Because the custom property is consumed inside `.item`'s transform, an invalid value invalidates the **whole declaration** — so every badge loses its centering and jumps down-right by half its size. Main's press-effect comment describes exactly this symptom: *"a bare scale() would drop that and jump the device down-right by half its size."*

Guard where the transform is built — `Number.isFinite(scale) ? … : IDENTITY_ZOOM` in `areaZoomTransform`, or run the value through `cssNumber` like every other config number reaching an inline style.

src/floorplan-card.ts
```diff
@@ -752,7 +821,10 @@ export class FloorplanCard extends LitElement {
     }
     .item {
       position: absolute;
-      transform: translate(-50%, -50%);
+      /* Counter-scaled against the zoom-to-room transform (--fp-inv-zoom,
+         set on .items) so a badge stays a constant, legible screen size
+         instead of ballooning with the room it's tapped into. */
+      transform: translate(-50%, -50%) scale(var(--fp-inv-zoom, 1));
```
nicosandller, line 1086: **This counter-scale is discarded by main's press effect once you rebase.**

`#142` restates the whole transform rather than composing onto it:

```css
.press-scale .item.interactive:active {
  transform: translate(-50%, -50%) scale(${PRESS_SCALE});
}
```

So at 4x zoom, pressing a device drops the 0.25 counter-scale and the badge jumps to roughly 4x its resting size for the length of the press — the opposite of the intended subtle shrink.

Multiplying instead keeps both working:

```css
transform: translate(-50%, -50%) scale(calc(var(--fp-inv-zoom, 1) * ${PRESS_SCALE}));
```

The same applies to any other rule that restates `.item`'s transform.

src/floorplan-card.ts
```diff
@@ -661,6 +696,40 @@ export class FloorplanCard extends LitElement {
...
+       HTML overlay so a transform here reframes both layers identically —
+       transform-origin:0 0 matches the translate-percent math in
+       areaZoomTransform(). Setting a transform (even the identity) makes this
+       div establish the containing block for its absolutely-positioned
+       svg/.items children, so it needs the same inset:0 they'd otherwise use. */
+    .plan-zoom {
+      position: absolute;
+      inset: 0;
+      transform-origin: 0 0;
+      transition: transform 0.4s ease;
```
nicosandller, line 949: The zoom animates over 400ms but the counter-scale doesn't: `--fp-inv-zoom` is an inline custom property that changes in one frame, and `.item` / `.text` / `.area-label` have no transition on `transform`.

So for the whole 400ms the two halves of the cancellation are out of step — tap a room that zooms to 4x and every badge and label snaps to a quarter size immediately, then the plan grows around them and they only look right once it lands. Giving the counter-scaled elements the same duration and easing as `.plan-zoom` keeps them cancelling throughout.

src/floorplan-card.ts
```diff
@@ -587,6 +609,17 @@ export class FloorplanCard extends LitElement {
               (it) => this._renderItem(it, c, rot)
             )}
           </div>
+          </div>
+          ${zoomedArea
```
nicosandller: Gated on `zoomedArea` rather than on the zoom actually doing something, so a tap that changes nothing still produces a zoom-out affordance.

Tap an area covering the whole canvas — a common "whole floor" polygon — and `areaZoomTransform` returns `scale 1, tx 0, ty 0`. Nothing moves, but the button appears, and clearing it needs a second tap on the button or the room. `${zoom.scale > 1 ? … : nothing}` matches what the user can actually see.

src/floorplan-card.ts
```diff
@@ -417,6 +426,12 @@ export class FloorplanCard extends LitElement {
           active.walls
         )
       : nothing;
+    // Zoom-to-room (tap an area). Both the SVG and the HTML overlay live
+    // inside one transformed wrapper below, so the two layers — positioned
+    // completely differently (a group transform vs. per-point left/top%) —
+    // reframe identically instead of drifting apart under zoom.
+    const zoomedArea = active.areas?.find((a) => a.id === this._zoomedAreaId);
+    const zoom = zoomedArea ? areaZoomTransform(zoomedArea.points, c.width, c.height, rot) : IDENTITY_ZOOM;
```
nicosandller, line 547: Context rather than a defect in this line: the branch is 8 commits behind main, and `git merge origin/main` conflicts in 5 hunks of this file.

The structural one is here — main now wraps the SVG in `keyed(c.skin ?? "", svg\`…\`)` so a skin change rebuilds the subtree, and `.plan-zoom` has to wrap both that and the `.items` div, so the opening `<div>` belongs outside the `keyed()` call. #148 also changed `_renderText` / `_renderAreaLabel` to take a `scale` argument and made `.area-label`'s font-size conditional, both on lines this PR touches.

Worth rebasing before the next round, so the diff shows what would actually ship.

nicosandller (changes requested):
## Requesting changes

Nice idea, and the hard part — one wrapper around both layers so the SVG and the HTML overlay reframe together — is the right architecture. `areaZoomTransform`'s rotation handling is correct too; I checked it against `rotatePlanPoint` by hand for all four angles and the test's reasoning holds.

I built the branch and drove it in a browser rather than reading the diff alone, which is where most of the below came from. Three things need fixing before this ships.

### Blocking

1. **[The pan is never clamped to the plan's edges](https://github.com/nicosandller/easy-floorplan/pull/158/files)** — zooming a room near a corner shows blank card where the plan should be. Measured 93px left / 36px top on a 698px-wide card.
2. **The pan still applies when the scale floors to 1** — tapping a room taller than the canvas slides the plan 192px sideways with *no zoom at all*. Same missing clamp; both fall out of one line.
3. **A single non-numeric area coordinate misplaces every device on the plan** — `--fp-inv-zoom:NaN` invalidates the whole `.item` transform, so the `translate(-50%,-50%)` centering goes with it.

### Worth fixing while you're in here

4. The press effect will discard the counter-scale once rebased (badge balloons ~4x on tap at 4x zoom).
5. The counter-scale snaps instantly while the plan animates for 400ms, so labels are visibly wrong-sized for the whole transition.
6. The zoom-out button appears even when the tap produced no zoom.

### Rebase first

The branch is **8 commits behind main** and `git merge origin/main` conflicts in 5 hunks of `floorplan-card.ts`. The structural one: main wraps the SVG in `keyed(c.skin ?? "", svg\`…\`)` for skin repaints, so `.plan-zoom`'s opening `<div>` has to sit outside that call. #148 also changed `_renderText` / `_renderAreaLabel` to take a `scale` parameter and made `.area-label`'s font-size conditional — both on lines this PR edits. Worth rebasing before the next round so the diff reflects what would actually ship.

### One note on reproducing these

If you verify in a hidden browser tab or a background pane, **CSS transitions are throttled and never advance**, so `.plan-zoom` reads as `matrix(1,0,0,1,0,0)` forever and every measurement silently reflects the un-zoomed state. I nearly filed a false "the transform never applies" bug on that. Force `transition: none` on `.plan-zoom` before measuring geometry.

obbers:
Rebased on main (the 5-hunk conflict was exactly what you flagged — `.plan-zoom` now wraps the `keyed(c.skin ?? "", …)` block rather than sitting inside it, and `_renderText`/`_renderAreaLabel` carry both the `#148` `scale` param and the counter-scale together) and fixed all three blocking issues plus the three follow-ups:

**Blocking**
1. **Pan clamp** — added the clamp you suggested almost verbatim: `Math.min(0, Math.max(100 * (1 - scale), t))` on both `txPercent`/`tyPercent`. Verified against your repro (canvas 980×700, area (40,40)-(140,110)) — clamps to (0, 0) instead of the previous 13.27%/7.14%.
2. **Same clamp fixes the scale-floors-to-1 case for free**, as you noted — added a dedicated test for your second repro too (area (40,40)-(400,660), scale floors to 1, pan now 0/0 instead of 27.55%).
3. **NaN guard** — `areaZoomTransform` now returns `IDENTITY_ZOOM` when `scale` isn't finite, checked right after the `Math.max(1, Math.min(maxScale, …))` line and before anything derived from it reaches a template. Added a test with one `NaN` coordinate.

**Follow-ups**
- **Press effect**: `.press-scale .item.interactive:active` now does `scale(calc(var(--fp-inv-zoom, 1) * ${PRESS_SCALE}))` instead of restating the transform bare — multiplies the counter-scale rather than dropping it. Fixed the matching reduced-motion rule too.
- **Transition sync**: `.item`, `.text`, and `.area-label` all got `transition: transform 0.4s ease` — same duration/easing as `.plan-zoom` — so the counter-scale animates in lockstep instead of snapping instantly.
- **Zoom-out button**: gated on `zoom.scale > 1` instead of `zoomedArea`, so tapping a whole-floor-sized area (scale stays 1) doesn't produce a dead affordance.

I don't have browser automation in my environment, so I measured all of this by reasoning through the transform math and mirroring your exact repro numbers in unit tests (`areaZoomTransform`'s test suite now covers: corner clamp, scale-floors-to-1 clamp, and the NaN guard) rather than by driving a live browser myself — worth an eyeball pass on your end given that. `npm run typecheck`, `npm test` (767 passing), and `npm run build` are all clean.

nicosandller (approved):
## Approving

All seven are fixed, and I re-measured each rather than reading the diff — the pan bugs only showed up as pixel geometry, so that is the only way to know.

### The three blockers

Same probes as last time, on `ac18349` with the transition neutralised:

| room | before | now |
| --- | --- | --- |
| tall (unzoomable) | `translate(27.551%, 0%) scale(1)` — 192px slide, no zoom | `translate(0%, 0%) scale(1)` — no pan |
| corner (4x) | 93px blank left, 36px blank top | **0px blank on all four sides** |
| malformed coordinate | `.item` transform `none` — every badge displaced | identity zoom, `--fp-inv-zoom: 1`, centering intact |

The clamp collapsing to `[0, 0]` at `scale === 1` is a neat way to make one line answer both pan findings. And guarding on `!Number.isFinite(scale)` fixes the NaN case at the source rather than at the style sink — worth noting that it holds for every route in, since any non-finite coordinate propagates through `bw`/`bh` into `scale` before it can reach `cxFrac`.

### The rest

- **Press effect** composes now — `scale(calc(var(--fp-inv-zoom, 1) * ${PRESS_SCALE}))`. You also caught the reduced-motion `:active` variant, which I had missed; it would have dropped the counter-scale the same way.
- **Zoom-out button** gated on `zoom.scale > 1`, so an unzoomable room no longer offers one. Verified: absent for `tall`, present for `corner`.
- **Rebased** onto current main, and `.plan-zoom` correctly wraps the `keyed()` block rather than sitting inside it, so a skin change rebuilding that subtree cannot disturb the zoom. That was the awkward part of the merge and it is resolved the right way round.

767 tests pass, typecheck and build clean.

### Two follow-ups, not blocking

Both are residue from the transition fix rather than anything new, and both are polish:

1. **The press rule outranks the new `.item` transition.** Measured on a default card (`pressEffect` defaults to `scale`): `.plan-zoom` is `transform 0.4s`, but every `.item` computes `transform 0.26s cubic-bezier(0.2, 0.8, 0.3, 1)` — `.press-scale .item.interactive` wins on specificity. So badges finish counter-scaling 140ms before the plan finishes zooming. Same desync the 0.4s was added to remove, just smaller. Labels and room names are fine; they compute the full `0.4s`.

2. **Reduced motion snaps the plan but animates the labels.** `.plan-zoom` gets `transition: none` under the preference, but `.text` and `.area-label` keep their 0.4s and the existing reduced-motion block only covers `.press-* .item.interactive`. So with reduced motion on, the plan jumps while every room name slides — animating for a user who asked not to, and inverting the desync.

Neither is worth another round. The feature is correct now, and it was the right call to fix the pan at the maths rather than patching symptoms.

PR #164, opened 2026-08-14
Trackers: initials can ride the dot, so people stay tellable apart

frido100:
## What

An optional `label` on a tracker draws short text (initials like `FR`) in the triangle's place — same class, so it keeps the pulse, the ripple rings and the glide between readings. A `paint-order: stroke` halo in `--fp-skin-card-bg` keeps the initials readable over busy plans without boxing them.

Empty or whitespace-only labels render the classic triangle, so existing configs are untouched.

## Why

We track four people across three floors. Each tracker has its own color, but with several dots on one floor the color legend stops carrying the "who" — you end up cross-referencing a legend in the margin. Initials on the dot answer it in place.

(We've been running exactly this as a local patch on the minified bundle since v1.1.x — upstreaming it so it survives updates.)

## How

- `Tracker.label?: string` in `types.ts`
- `renderTracker`: 2-sensor branch swaps `<polygon>` for `<text class="tracker-dot">` when a trimmed label is present (1-sensor line mode unchanged — there's no dot to label)
- `trackerForm`: a plain text field, same pattern as the area name
- README: editor section + `Tracker` reference
- 3 tests in `render.test.ts` (triangle without label, text with label, whitespace-only falls back)

`vitest` 957/957 green, `tsc --noEmit` clean, `npm run build` ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,8 +281,10 @@ rings; readings outside `[min, max]` clamp to the rectangle's edge. With one, a 
 pulsating line spans the unknown axis — honest about knowing only one coordinate. With
 neither reporting, nothing renders.
 
-The rectangle itself is editor-only; the card shows just the marker. **Color** and **dot
-size** are per tracker.
+The rectangle itself is editor-only; the card shows just the marker. **Color**, **dot
+size** and **label** are per tracker — a label (initials like `FR`) rides in the
+triangle's place with the same pulse and glide, so several people on one floor stay
+tellable apart when color alone isn't enough.
 
 #### Presence ripples
 
@@ -573,7 +575,7 @@ A live (x, y) position estimate driven by one or two orthogonal distance sensors
 animated inside a rectangular tracked area:
 
 ```yaml
-{ id, x, y, w, h, angle?, color?, dotSize?,
+{ id, x, y, w, h, angle?, color?, dotSize?, label?,
   xSensor?: { entity, min, max, invert?, presence?: { entity, invert? } },
   ySensor?: { entity, min, max, invert?, presence?: { entity, invert? } } }
 ```
@@ -583,6 +585,8 @@ animated inside a rectangular tracked area:
   `[min, max]` onto the rectangle's edges along that axis; `invert` flips the mapping.
 - `presence` — a binary gate per axis; if either reports clear (or `unavailable` /
   `unknown`) the marker hides. `invert` flips on/off, never the unavailable case.
+- `label` — short text (initials) drawn in place of the triangle, keeping its pulse,
+  ripples and glide. Empty / unset keeps the classic triangle.
 - Both sensors → a pulsating triangle with ripple rings. One → a faint pulsating line
   across the unknown axis. The rectangle is editor-only.
 
--- a/src/editor-forms.ts
+++ b/src/editor-forms.ts
@@ -863,6 +863,8 @@ export function trackerForm(tr: Tracker): FormSpec {
         label: "Dot size",
         selector: { number: { min: 6, max: 80, mode: "slider", unit_of_measurement: "px" } },
       },
+      // Initials in place of the triangle — see Tracker.label.
+      { name: "label", label: "Label", selector: { text: {} } },
     ],
     data: {
       w: tr.w,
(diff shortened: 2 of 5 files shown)


nicosandller:
@frido100  thanks for the feedback and opening a PR. 

Im really curious at how you're using trackers. So allow me to ask a few questions questions!


You are tracking x number of people across x different floors correct? And you're using X*2 number of distance tracking sensors for this. 

Yet you have put all the tracker entities on the same floor of the floorplan so you need to change the arrows in the tracker for initials to differentiate them?

Did I get the situation correctly?

Can I ask:

- why are the trackers in the same `floor`  and not in different floors? 
- is this a family house or an office?
- why aren't colors enough to differentiate?

If possible sharing a screenshot would help imagine this better.


I'm asking so I understand the direction of this plugin as a product better to reconcile this with other use cases. Thank you!

amadeobriones:
Not weighing in on whether the feature lands — that's the conversation above. But in case it does,
I run a fork of this card and have been testing the open PRs against current `main`, and this one
merges cleanly and type-checks but fails one test:

```
every field lands in exactly one panel group > covers every furniture, tracker and area field
expected { what: 'tracker', ungrouped: ['label'] } to deeply equal { what: 'tracker', ungrouped: [] }
```

The branch is ~49 commits behind, so it predates the collapsible panel groups. The tracker panel
now renders only through `formSlice(trSpec, TRACKER_GROUPS[n][1])` (`editor.ts:4792` and `:4804`),
and `trSpec` isn't used anywhere else — so a field named in neither group is never rendered. I
checked that directly rather than inferring it: `trackerForm(tr).fields` does contain `label`,
while the union of both rendered slices does not. **After a rebase the new option wouldn't appear
in the editor at all.**

Two lines fix it, and the suite goes green (1171/1171):

```diff
# src/editor.ts:2067
-    ["Marker", ["dotSize"]],
+    ["Marker", ["dotSize", "label"]],
```
```diff
# src/editor-forms.test.ts:1734
-  const TRACKER_GROUPS = [["w", "h", "x", "y", "angle"], ["dotSize"]];
+  const TRACKER_GROUPS = [["w", "h", "x", "y", "angle"], ["dotSize", "label"]];
```

Both are needed — the test keeps its own copy of the group lists, since the real ones are
`private static` on the editor class.

Two small things, take them or leave them:

- `label` isn't length-capped. At `dotR * 1.6` a long string spreads well past the dot, and most
  other user-supplied values here get clamped.
- `dominant-baseline="central"` has historically been uneven in some WebKit versions — might be
  worth a look in the iOS companion app before merging.

Separately: the Codex note about rotated plans reproduces. Trackers are rendered inside the
`<g transform=${rotTransform}>` group (`floorplan-card.ts:886`, and the `renderTracker` call at
`:1132` is still inside it), whereas badges and labels are HTML anchors remapped through
`rotatePlanPoint` so they stay upright — the comment at `:636` says exactly that. So on a plan with
`rotation: 90/180/270` the initials would turn with the plan where every other piece of text
doesn't. Worth folding in if this goes ahead.

nicosandller (changes requested):
One thing to fix, and it's only because `main` moved under you.

### The Label field won't show up in the editor after a rebase

This branch is from before the element panel was split into titled groups. The panel now renders each group by slicing the form (`editor.ts` → `TRACKER_GROUPS`), so a field that isn't in a group is never drawn — you added `label` to `trackerForm` but not to the group list, so the control silently disappears. YAML would still work; the editor wouldn't.

`main` has a test that catches exactly this. Merged with current `main`:

```
× every field lands in exactly one panel group > covers every furniture, tracker and area field
  ungrouped: [ "label" ]
```

The fix is one line in `src/editor.ts`:

```js
private static readonly TRACKER_GROUPS = [
  ["Zone", ["w", "h", "x", "y", "angle"]],
  ["Marker", ["dotSize", "label"]],
] as const;
```

So: merge `main` in, add `label` to the `Marker` group, and the suite goes green (1179 tests). On the branch as it stands the suite is green too (957) — it's only the rebase that surfaces this.

### One optional thought

`stroke-width="3"` on the halo is a fixed number in user units while the font size scales with `dotR`. At `dotSize: 40` the halo will look thin, at `dotSize: 8` it'll swallow the letters. `stroke-width=${dotR * 0.25}` or similar would keep it proportional. Not blocking — your call.

Nothing else from me. Once the rebase is in I'm happy to merge this.

---

nicosandller:
Closing this pull request for a lack of response.

PR #167, opened 2026-08-15
fix(devices): a press lands on the device, not on the air its ripple fills

Gifford47:
### Problem

A presence device with a ripple is a hit target the size of its ring. `.item` takes pointer events for its whole box, and the box grows to hold the decoration, so a 30px motion icon with `rippleSize: 110` answers presses across 110 square pixels — as a *square*, so even the corners outside the visible ring count.

That is wrong twice over. It makes a small icon behave like a large button, and since the overlay sits above the plan it swallows presses meant for whatever is underneath: a door, a device standing inside the ring, or in the editor the wall you are trying to select.

The ring is not a control. It says "presence here"; the badge is what you press.

### Change

The anchor stops taking pointer events, and the parts that *are* the device take them back — the badge and its label. `.stack-icon` goes with the decoration, since it spans the whole ripple (`inset: 0`) and would otherwise keep the old box.

A ripple-only device draws no badge to aim at, so its centre answers instead: a `MIN_TOUCH_TARGET` disc on the ripple's dot, invisible, 34px. Deliberately fixed and **not** scaled by `overlayScale` (#148) — a minimum touch target is a fact about fingers, and those do not shrink with the plan.

The editor gets the same rule, so what you can grab there is what you can press on the card, and the wall under a ripple is selectable again.

### Notes

**Nothing else changes.** A device without a ripple is unaffected: its badge was always the whole of its box.

**Not unit-testable here.** This is CSS, and the suite covers pure functions only — `npx tsc --noEmit` is clean and all tests pass, but none of them speak to this. Verified by hand: with a ripple device on the plan, a press just outside the badge but inside the ring now reaches the plan underneath.

### DIFF
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -19,6 +19,7 @@ import {
   DEFAULT_TEXT_SIZE,
   DEFAULT_RIPPLE_SIZE,
   DEFAULT_AREA_LABEL_SIZE,
+  MIN_TOUCH_TARGET,
   DEFAULT_SUN_MIN,
   DEFAULT_SUN_MAX,
   PRESS_SCALE,
@@ -1272,6 +1273,15 @@ export class FloorplanCard extends LitElement {
       inset: 0;
       pointer-events: none;
     }
+    /* A device's hit area is what you can *see* of it — the badge and its
+       label — not the box its decoration happens to fill.
+
+       A presence ripple is 80–110px of mostly empty air, and the anchor grew
+       to hold it: a 30px motion icon behaved like a 110px square button, which
+       also swallowed taps meant for the plan underneath it. The ring is
+       decoration; it says "presence here", it is not a control. So the anchor
+       stops taking pointer events and the parts that are the device take them
+       back. */
     .item {
       position: absolute;
       /* Counter-scaled against the zoom-to-room transform (--fp-inv-zoom,
@@ -1283,13 +1293,44 @@ export class FloorplanCard extends LitElement {
          every badge is briefly the wrong size mid-transition. */
       transform: translate(-50%, -50%) scale(var(--fp-inv-zoom, 1));
       transition: transform 0.4s ease;
-      pointer-events: auto;
+      pointer-events: none;
       /* Not a hand: a device with nothing bound, or tap_action set to none,
          is not a button (issue #134). Only .interactive gets the pointer. */
       cursor: default;
       display: flex;
       flex-direction: column;
       align-items: center;
+    }
+    .item .badge,
+    .item .label {
+      pointer-events: auto;
+    }
+    /* .stack-icon spans the whole ripple (inset: 0), so it has to stay out of
+       the way too — the badge inside it is the target, not its wrapper. */
+    .stack-icon,
+    .ripple,
+    .press-ink {
+      pointer-events: none;
+    }
+    /* A ripple-only device draws no badge, so its centre has to answer for it,
+       or switching the badge off would leave the device unclickable. The dot
+       is 8px across; this gives it a real touch target without drawing one.
+       Deliberately a fixed size, and not scaled by overlayScale (#148): a
+       minimum touch target is about fingers, which do not shrink with the
+       plan. */
+    .item.interactive .ripple .dot {
+      pointer-events: auto;
+      position: relative;
+    }
+    .item.interactive .ripple .dot::after {
+      content: "";
+      position: absolute;
+      left: 50%;
+      top: 50%;
+      width: ${MIN_TOUCH_TARGET}px;
+      height: ${MIN_TOUCH_TARGET}px;
+      transform: translate(-50%, -50%);
+      border-radius: 50%;
     }
     .item.interactive {
       cursor: pointer;
--- a/src/types.ts
+++ b/src/types.ts
@@ -852,6 +852,15 @@ export const FURNITURE_GLOW_TRANSMISSION = 0.5;
 export const DEFAULT_TRACKER_DOT_SIZE = 14;
 
 export const DEFAULT_ITEM_SIZE = 34;
+/**
+ * Smallest area that answers a press, in screen pixels — for a device whose
+ * only visual is a presence ripple and so has no badge to aim at.
+ *
+ * Screen pixels on purpose, and never scaled by `overlayScale`: this is a
+ * measure of fingers, not of the drawing. A plan shrunk into a narrow card
+ * would otherwise shrink its touch targets with it.
+ */
+export const MIN_TOUCH_TARGET = 34;
 export const DEFAULT_TEXT_SIZE = 16;
 export const DEFAULT_RIPPLE_SIZE = 80;
 /** Neutral gray, so furniture reads differently from the walls. Skinnable (#122). */
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -32,6 +32,7 @@ import {
   DEFAULT_CUSTOM_PERCENT,
   DEFAULT_GRID,
   DEFAULT_ITEM_SIZE,
+  MIN_TOUCH_TARGET,
   DEFAULT_TEXT_SIZE,
   DEFAULT_WIDTH,
   DEFAULT_HEIGHT,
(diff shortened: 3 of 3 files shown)


src/types.ts
```diff
@@ -852,6 +852,15 @@ export const FURNITURE_GLOW_TRANSMISSION = 0.5;
 export const DEFAULT_TRACKER_DOT_SIZE = 14;
 
 export const DEFAULT_ITEM_SIZE = 34;
+/**
+ * Smallest area that answers a press, in screen pixels — for a device whose
+ * only visual is a presence ripple and so has no badge to aim at.
+ *
+ * Screen pixels on purpose, and never scaled by `overlayScale`: this is a
+ * measure of fingers, not of the drawing. A plan shrunk into a narrow card
+ * would otherwise shrink its touch targets with it.
```
nicosandller, line 861: Non-blocking, naming rather than value: `MIN_TOUCH_TARGET = 34` is below every published minimum — WCAG 2.5.8 sets the AA floor at 24×24 and AAA at 44×44, and Material and the Apple HIG both say 44. The 34 is `DEFAULT_ITEM_SIZE`, i.e. "the size of one badge", which is a reasonable choice here (a larger target covers more of the plan, which is the problem this PR is reducing) — but a constant *named* `MIN_TOUCH_TARGET` reads as a conformance claim to whoever finds it next.

`RIPPLE_HIT_SIZE` would say what it actually is. Or keep the name and raise it to 44, accepting slightly more plan coverage. Either is fine; the mismatch between the two is the only thing I'd change.

src/floorplan-card.ts
```diff
@@ -1283,14 +1293,45 @@ export class FloorplanCard extends LitElement {
...
+      pointer-events: none;
+    }
+    /* A ripple-only device draws no badge, so its centre has to answer for it,
+       or switching the badge off would leave the device unclickable. The dot
+       is 8px across; this gives it a real touch target without drawing one.
+       Deliberately a fixed size, and not scaled by overlayScale (#148): a
+       minimum touch target is about fingers, which do not shrink with the
+       plan. */
+    .item.interactive .ripple .dot {
+      pointer-events: auto;
```
nicosandller, line 1322: The reasoning here is right and I'd keep the behaviour — but the tradeoff has a limit worth writing down, because the comment currently reads as unconditional.

Measured: a 240px-wide card with `overlayScale: "plan"` renders a `rippleSize: 110` ring at **33px** on screen, against this fixed **34px** target — so the target is already marginally larger than the ring it stands in for. A 2000-unit canvas on a phone-width card puts the ring near 13px against the same 34px.

Still a 3× improvement on the 55px it replaces, and scaling the target with the plan would defeat the point of having one. I'd just bound the claim, so nobody reopens this later thinking the overshoot was unnoticed.

nicosandller (approved):
Approving. The fix is right, the reasoning in the description is sound, and I could not break it.

Since this is CSS and the suite covers pure functions only, I measured it rather than read it — hit-testing with `elementFromPoint` across the shadow boundaries, plus real clicks and a real drag. (Note for anyone reproducing: the preview pane reports a 0×0 viewport while hidden, so `elementFromPoint` returns `null` for everything and `requestAnimationFrame` never fires. Size the viewport first or every probe silently reads as "nothing".)

### Distance from centre at which a press stops reaching the device

| device | before | after |
|---|---|---|
| `iconRipple`, ring 110, badge 30 | 55px, **square** | 17px |
| `ripple` only, ring 110 | 55px, square | 17px |
| plain badge, no ripple (control) | 18px | 18px — unchanged |

From 18px out, the deepest hit is the plan: `<line.wall-hit>` in the editor and `<line>` on the card. So the wall under a ripple really is selectable again, and the editor and the card agree to the pixel — which was the other thing you claimed.

### What I tried to break, and couldn't

- **The press effect.** `pointer-events: none` on the anchor could plausibly have stopped `.press-scale .item.interactive:active` matching. It doesn't — on a real click, a `mousedown` listener on `.item` observes `target: badge`, `currentTarget === .item`, and `item.matches(':active') === true`. `:active` still walks the ancestor chain of the actual hit target.
- **The gesture path.** `action-handler` binds `mousedown`/`touchstart`/`click`/`keydown` on `.item` and uses no pointer capture and no `e.target` checks, so bubbling covers it. Confirmed end to end: click on the badge → `click` reaches `.item` → `hass-more-info` for the right entity → `action{tap}`. Clicking the ring's empty air produces an entirely empty log.
- **`_startInk`.** Reads `ev.currentTarget` and that element's rect, so the ripple press effect still positions from the anchor even though the press lands on a descendant.
- **Editor drag.** Grabbed the badge and dragged: the item moved 200,200 → 260,160 and selected. Dragging is by the badge now, which is the point.
- **Selector reach.** `.stack-icon`, `.ripple` and `.press-ink` are unscoped, so I checked what else they could hit — nothing. `renderRipple` has exactly four call sites, all devices, and the tracker uses `tracker-dot`.

One detail worth calling out because it's easy to get wrong and you got it right: `border-radius: 50%` genuinely clips hit-testing, so the target is a real disc, not a square wearing a rounded corner. 16px on-axis hits; 16px on both axes (distance 22.6) misses; 11/11 (15.6) hits. "A `MIN_TOUCH_TARGET` disc" is literally accurate.

954 tests, `typecheck` and `build` all clean, and the branch is level with `main`.

---

### Three things for later, none blocking

**1. The constant is below every published minimum, and the name promises otherwise.** `MIN_TOUCH_TARGET = 34` is `DEFAULT_ITEM_SIZE` — "one badge" — but WCAG 2.5.8 puts the AA floor at 24×24 and AAA at 44×44, and Material and the Apple HIG both say 44. A constant *named* `MIN_TOUCH_TARGET` sitting at 34 will read as a standard to the next person.

I don't think the number is wrong: a bigger target covers more plan, which is the thing this PR exists to reduce, so 34 is a fair balance. It's the name I'd change — `RIPPLE_HIT_SIZE` says what it is without claiming conformance. Your call either way.

**2. The `overlayScale` tradeoff has a limit the comment doesn't mention.** "A minimum touch target is about fingers, which do not shrink with the plan" is the right call. But the target can end up larger than the ring it stands in for: on a 240px-wide card with `overlayScale: "plan"`, a `rippleSize: 110` ring measures **33px** on screen against the fixed **34px** target. Push the canvas to 2000 units on a phone-width card and the ring lands near 13px against the same 34px.

That is still a 3× improvement on the 55px it replaces, and scaling the target would defeat the purpose — so I'd keep the behaviour and just bound the claim in the comment, so nobody re-opens this thinking the overshoot was an oversight.

**3. "Nothing else changes" has one exception.** An inert ripple device — `tap_action: none`, so no `.interactive` — is now fully transparent: the dot rule is gated on `.interactive`, and I measured the centre hitting `<line.wall>` behind it. That is correct per #134 and a bonus (an inert ring stops blocking the plan), but it also means its `title` tooltip is unreachable by hover, and for a ripple-only device with no label that was the only way to identify it on the card. Worth a sentence in the description rather than a code change.

On the untestable point — agreed, and the repo has precedent for the next best thing: `FURNITURE_GLOW_TRANSMISSION` has a test asserting it stays strictly between 0 and 1, precisely because it's a tuned constant no behavioural test covers. A two-line guard on `MIN_TOUCH_TARGET` would fit that pattern.

nicosandller (approved):
Duplicate of my earlier review above — posted without checking that one existed. Trimming it to the one thing it adds, so it isn't a wall of repeated text.

Still approving; everything in the first review stands, and I reproduced it independently (the before/after hit distances, taps firing from the bare dot, the editor reaching `line.wall-hit` again, and `position: relative` on `.dot` leaving its centre at dx 0 / dy 0).

**The one new note: two of the three `pointer-events: none` declarations are redundant, and the comment explains them wrongly.**

The comment says `.stack-icon` "spans the whole ripple (`inset: 0`), so it has to stay out of the way too". But `inset: 0` grows the *box*, and box size has nothing to do with hit-testing eligibility — a descendant of a `none` anchor is already not a target, whatever its size. I deleted the rule from the live stylesheet rather than argue it:

```
deleted: .stack-icon, .ripple, .press-ink { pointer-events: none; }
computed after deletion:  .stack-icon -> none    .ripple -> none
hit-testing after deletion: identical at every sampled point
```

They inherit `none` from `.item`, and none of them set `auto` before this PR. Harmless, and defensible as belt-and-braces in case the anchor ever takes events again — but as written the comment will send a future reader after the wrong mechanism. Worth either dropping the declarations or saying that's why they're there.

Issue #73, opened 2026-07-14
third mode for windows


gorischek:
Windows currently only have two modes: slide and swing. The problem is that "swing" shows two sashes, but we have only one sash per window. As a workaround I'm using doors as icons for windows.
So if possibel I would ask for:
slide, single sash, double sash

gorischek:
thanks a lot!

PR #168, opened 2026-08-15
feat(openings): two leaves, for the doors that have two

Gifford47:
### Problem

A window can say how many sashes it has; a door cannot. Whatever its width, a door is drawn as one leaf spanning the whole opening — so a double door, a French door, a pair of patio leaves all come out as a single slab sweeping an arc twice as wide as any of them really does.

`sash` existed already, but `windowSash()` only read it for windows and told doors they were `"double"`, which was harmless only because the door branch ignored it.

### Change

`sash` now reads on both types, with the default per **type** rather than per field:

| type | default | the exception you write down |
| --- | --- | --- |
| window | `double` — two casement leaves | `single` (issue #73) |
| door | `single` — one leaf | `double` |

Both defaults are the ordinary case for their own type, so existing configs are untouched and neither default lands in the YAML.

A double door keeps the door's own symbol: two leaves and two arcs, **no jambs**. Jambs are what say "glass" and are the whole visual difference from a casement pair. Each leaf is half the opening and its arc has half the radius, because that is the distance that leaf's tip actually sweeps. The animation rides the existing `fp-door-leaf` / `fp-leaf-r` classes, so it opens with everything else.

Only swing motion: a slider or a roller has no hinged leaves, so it reports its type's default and nothing reads it.

**Editor.** The field is offered on doors too, labelled **Leaves** there and **Sashes** on windows, with the options ordered so each type's default comes first. **Hinge** now appears only where there is one leaf to hang — a double is hinged at both jambs, so there is no side left to choose.

**Refactor that came with it.** The three swing branches in `renderOpening` were three near-copies of the same markup, and that is precisely how the door came to have no way of being a double. They are one branch now, deciding the two things that actually differ: the leaf count and the jambs. `windowSash` is superseded by `openingSash`, which answers the same question for both types.

### Notes

`npx tsc --noEmit` clean, **966 tests** passing, 12 of them new: the two-leaf geometry and its arcs, the jamb difference against the casement window it mirrors, the untouched single door, the sliders and rollers that ignore the setting, and the per-type patch round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,7 +389,7 @@ distorted anyway.
 | `id`          | string                      | Unique id.                                             |
 | `type`        | `door` \| `window`          | The kind of opening.                                   |
 | `motion`      | `swing` \| `slide` \| `roll` | How it moves: hinged (default), sliding panels, or a roll-up curtain (garage / roller shutter). |
-| `sash`        | `single` \| `double`        | Swing windows only: one full-width sash, or the classic two. Default `double`. |
+| `sash`        | `single` \| `double`        | Swing openings only: how many hinged leaves. The default differs by type, because the ordinary cases do — a window opens with `double` (two casement sashes), a door with `single` (one leaf across the opening). Set it to draw a single-sash window or a **double door**; both leaves then hinge at their own jamb and trace their own arc. Ignored by sliding and rolling openings. |
 | `shutterEntity` | string                     | An external shutter over the same gap (`cover` or contact), with its own open/closed state. With `entity` bound too, the card draws the shutter's own icon beside the opening — open/closed in both glyph and colour — and tapping that icon opens the shutter. |
 | `shutterStyle` | `swing` \| `roll`           | Louvered panels or a roll-up curtain. Defaults from the entity (contact → `swing`, `cover` → `roll`). |
 | `shutterInvert` | boolean                   | Flip the shutter's open/closed reading — a reed contact on hinged panels often reads `on` when they are shut. Separate from `invert`. |
--- a/src/editor-forms.test.ts
+++ b/src/editor-forms.test.ts
@@ -819,17 +819,45 @@ describe("wallForm / projectForm / floorImageForm", () => {
 describe("openingForm — sash and shutter (issues #73 / #74)", () => {
   const win = { id: "w1", type: "window", x: 0, y: 0, length: 90, angle: 0 } as Opening;
 
-  it("swing windows offer Sashes; single sash reveals Hinge", () => {
+  it("every swing opening offers a leaf count; only sliders and rollers don't", () => {
     const names = openingForm(win).fields.map((x) => x.name);
     expect(names).toContain("sash");
-    expect(names).not.toContain("hinge"); // double: no single hinge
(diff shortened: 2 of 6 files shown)


nicosandller (approved):
Approving. `tsc` clean, **966 tests**, build clean — and I checked the risky part of this, which is the refactor rather than the feature.

## The branch collapse is provably behaviour-preserving

Three near-identical markup branches becoming one is exactly where a silent regression hides, so I didn't take the reasoning on trust. I rendered `renderOpening` for 13 opening configurations at three open amounts each, on this branch and on the base commit, normalised away comments and attribute quoting, and diffed the geometry.

The diff is confined to **one** region:

```
### window swing default (two sashes)   identical
### window swing single (#73)           identical
### window swing sash:double explicit   identical
### door swing default                  identical
### door swing sash:single explicit     identical
### door swing sash:double (NEW)     <- the only difference, at all three amounts
### door swing flipH                    identical
### door slide / slide sash:double      identical
### door roll                           identical
### window slide / roll / slide+sash    identical
```

Every pre-existing configuration emits byte-identical geometry, including sliding and rolling openings carrying a stray `sash` that nothing should read. The only behaviour that changed is the one that previously had no expression at all.

## The "no jambs" call is the right one, and not a new convention

I wondered whether a double door would be distinguishable from a casement pair at plan scale, since two jamb ticks is a small difference. It turns out to be exactly the same difference that already ships:

```
window swing single (#73)  vs  door swing default
  only in the window: two jamb lines at ±half, and the fp-opening-window class
  (leaf width and arc radius identical)

window swing default       vs  door swing sash:double (NEW)
  only in the window: the same two jamb lines, and the same class
  (leaf widths and both arc radii identical)
```

So the new pair inherits the single/door distinction the codebase already established for #73 rather than inventing a weaker one. Rendering the four symbols side by side across the open sweep, the double door reads correctly: two half-width leaves meeting in the middle, each tracing its own half-radius arc, animating together on the existing classes.

Per-type defaults are the right shape too — both defaults are their type's ordinary case, so neither lands in the YAML, and `toPatch` writing down only the non-default is what keeps that true in both directions. The tests cover the round-trip in both directions, the hinge gating, and the labels, and they're substantive rather than shape-checks.

## One nit, non-blocking

Turning a single door into a double hides the **Hinge** field, which is correct — but a door that already had `flipH: true` keeps it, and it is then unreachable from the editor. Harmless in practice: I confirmed the double symbol is mirror-symmetric, so the stranded `scale(-1 1)` changes nothing on screen at any open amount. It's just a key sitting in the YAML meaning nothing, which the neighbouring `toPatch` branches already tidy up (the slider style one clears `sliderStyle` and `secondaryEntity`). Clearing `flipH` when `sash` becomes `double` would close it.

Tiny second one, take it or leave it: the window helper still reads "Single = one full-width sash (issue #73)" while the new door helper describes the thing without an issue number. The door's phrasing is the better one for an end user.

Neither blocks. Nice refactor — the fact that the door had no way of being a double *because* the markup was copied three times is the clearest possible argument for merging the branches, and it's good that the commit message says so.

PR #171, opened 2026-08-16
Dev: replace the mock harness with a real Home Assistant in a container

nicosandller:
The only way to test this card was `dev/`, and `dev/` handed it a `hass` object this repo
wrote. Fast, and fine for laying out a plan, but it means the card could only ever be
checked against our own idea of Home Assistant. The places that idea is thinnest are
exactly the ones that break in the field: history and recorder behaviour, entities that go
`unavailable`, attributes that moved between HA versions, resource loading, theming.

So this replaces it with a real one, in a container, with a house that keeps moving.

```bash
npm run ha
```

Builds the card, registers it as a Lovelace resource, seeds an editable demo dashboard, and
starts Home Assistant on `localhost:8123` with `dist/` mounted at `/local`. Alongside it:
`ha:logs`, `ha:down`, `ha:reseed` and `ha:reset`.

## How to review this

**1. Start it.** With any Docker runtime running (`brew install colima docker
docker-compose && colima start` if you have none — Compose **v2** is required, since the
scripts call `docker compose`, not `docker-compose`):

```bash
npm run ha
```

First boot pulls the HA image and takes a couple of minutes.

**2. Do the onboarding once.** It ends on HA's "Create my smart home" screen. Any username
and password — the port is published on `127.0.0.1` only, so nothing is exposed to your
network. Skip location, analytics and the discovered-devices page. Only on the first run;
the account persists in `docker/config/.storage`.

**3. Open "Floorplan Demo" in the sidebar.** The sample plan, rendering live: lights
toggling and recolouring every 10s, light pools mixing where the two glow radii overlap, a
window at partial positions, the garage door rolling shut, a tracker drifting around the
room. Give the generators ~2 minutes to build up history before judging it.

**4. Edit it.** Click the pencil — the card's visual editor opens on the demo plan itself.
Drag a light somewhere, resize a wall, add a room. That dashboard is storage-mode, so
everything the editor does persists, and `src/editor.ts` is what you are exercising.

The plan still lives in git: `docker/config/floorplan-demo.yaml` is the committed source,
seeded into the dashboard on first run. Seeding only applies when the dashboard is absent,
so restarts never eat your work; `npm run ha:reseed` puts the committed version back when
an experiment wanders off.

**5. Watch it disagree with you.** The point of the whole thing. The **History** view (in
the same dashboard) has a graph over the same entities plus switches for the generators.
Two worth trying:

- Flip `input_boolean.history_generator` off — the plan goes still, useful when you want to
  read one state instead of chasing it.
- Watch `sensor.flaky_sensor`. Every five minutes it goes genuinely `unavailable` for 45
  seconds — not "off", not empty — and the badge has to survive it. That state is all over
  real installations and the old mock could not produce it.

**6. Iterate on the code.** `npm run watch` in a second terminal; `dist/` is mounted, so a
save lands in the container immediately. The browser has cached the old file, so
hard-refresh (Cmd/Ctrl-Shift-R). This is the one place the container is worse than the old
harness.

## Two things that look like bugs and are not

**Four pending updates, returning after every restart.** They are fake. The `demo`
integration ships six `update` entities that report "update available" forever, and demo
state is not persisted. They cannot be dismissed: `update.skip` clears two, and the other
four declare `auto_update: true`, which HA refuses to skip by design. Nothing needs
installing. It is the price of demo's lights carrying real `brightness`/`rgb_color` and
covers carrying a real `current_position`.

**One instance at a time.** The container name is fixed and Compose names its project after
the `docker/` directory, which exists in every checkout of this repo. A second worktree
running `npm run ha` would adopt the existing container — still mounting the first
checkout's config and `dist`. `prepare.mjs` refuses to start in that case and prints the
command to clear it. (This one cost me an evening of editing a config the running instance
was not reading, which is why the guard exists.)

## What this removes

`dev/` is gone — all four files, the `serve` script, and the README sections. Nothing in
`src/` imported it, so the deletion is clean; `npm test` (954) and `npm run typecheck` are
green without it.

**One real loss worth an opinion:** `/dev/symbols.html` drew every furniture symbol on one
page, and `furniture/README.md` pointed contributors at it for checking a new glyph at size
and stretched. There is no contact sheet any more — that doc now points at the editor,
which checks one symbol at a time. If you would rather keep it, it is the one piece of
`dev/` worth porting, and it never needed the `hass` mock.

## Choices worth a second opinion

**A seeded storage dashboard, not a yaml one.** A yaml-mode dashboard keeps the plan in git
but Home Assistant refuses to edit yaml dashboards in the UI — which, for a card whose
selling point is a drag-and-drop editor, is the wrong half to give up. A storage dashboard
is just two json files, so `prepare.mjs` writes them from the committed yaml: git keeps the
readable source, the instance gets an ordinary editable dashboard. Going the other way,
editor back to git, is a deliberate copy out of the raw configuration editor rather than
anything automatic.

**A Lovelace resource, not `frontend.extra_module_url`.** The YAML shortcut looks right and
is subtly wrong: it emits `<script>import("/local/easy-floorplan-card.js");</script>` into
the app shell, a bare dynamic import that nothing awaits. Whether the element is defined
before Lovelace renders the view is then a race decided by cache state and machine speed,
and losing it draws a bare "Configuration error" card instead of the plan. Resources are
fetched and awaited before dashboards render, and are how a real install loads the card.

**History only builds forward from boot.** Recorder timestamps are wall-clock. Handing
yourself a plan that was busy yesterday would mean writing recorder rows by hand, and that
schema moves between HA versions — you would end up debugging the fixture instead of the
card. So the generators make ~2 minutes enough to have something dense to scrub.

**Entities come from the `demo` integration**, because it provides the attributes the card
actually reads, which are tedious to reproduce by hand. Sensors, the door contact and the
tracker's distance readings are template entities over input helpers, so the generators have
something to drive. Demo ids can drift between HA versions; the list the sample plan uses is
at the top of `docker/config/floorplan-demo.yaml`.

## Verification

Booted and driven, not just written. HA's own `check_config` passes; every entity id in the
sample plan exists in the running instance; the card serves at `/local/easy-floorplan-card.js`
as `200`, `text/javascript`; a 15-minute run produced 4278 recorded state rows and 5 clean
`unavailable` cycles with zero errors or warnings in the log. The port binding was checked
from the LAN address (refused) and loopback (serves). The card itself was loaded into a DOM
the way the frontend loads it and handed this exact dashboard config — it registers, accepts
the config and renders. The seeded dashboard was booted on a scratch instance: Home
Assistant accepts it, reports it as `mode: storage`, and leaves the files intact.

Running it is what caught most of the bugs in the history: the demo covers are not
interchangeable (`garage_door` is OPEN|CLOSE only and threw every 30s under
`set_cover_position`), the tracker automation blocked itself by waiting 20s inside something
triggered every 5s, and the `extra_module_url` race above. None were visible from reading
the config.

**Not verified:** the browser-side render and the edit pencil, on the seeded dashboard.
Confirming those needs a logged-in session, and creating the account is a step I leave to a
human. Worth a look on your first run.


nicosandller:
Adding some pictures that illustrate what the dev instance looks like for testing:


Screenshot 2026-08-17 at 9 32 05 AM
Screenshot 2026-08-17 at 9 56 29 AM

nicosandller:
I want to know what @TruthOf42 / @superflyer11 / @shauneccles 

think of this way to test things! Please give it a try.

PR #174, opened 2026-08-18
A sensor per leaf, offline devices that read as offline, and a header worth its height (closes #159, #162, #152)

### LINKED ISSUE #152: [FR] reduce waste of screen space
When both project name defined and more Floors created -  lot of space on top of screen is actually wasted.

Image

Good news: project name can be set empty and one huge empty line disappears.
But Floor selector produce another empty region. And only 2 Floors now...

### LINKED ISSUE #159: Two sensors for the other two-part openings: double-sash windows, hinged shutters
Follow-up to #145 / #151. **Not needed for those to land** — this is the generalization, deliberately split out.

## What #151 established

It gives a sliding opening a sensor per panel, so a two-leaf patio door can draw left-open / right-shut instead of forcing both leaves onto one state.

The useful part is that the plumbing it added is **not slider-shaped**:

| piece | slider-specific? |
| --- | --- |
| `Opening.secondaryEntity` | no — a plain config key |
| `secondPanelOf(o)` → `{ ...o, entity: o.secondaryEntity }` | no |
| `OpeningStyle.second?: { amount, active }` | no |
| card's `_openingSecond(o)` | only via the gate below |
| `openingHasTwoPanels(o)` → `sliderStyleHasTwoPanels(...)` | **yes** — the gate |
| which render branch reads `amt2` / `tone2` | **yes** |

So the second sensor already flows end to end for any opening. What is slider-only is *the predicate that decides an opening has two parts*, and *which drawing bothers to read the second amount*.

## What is still stuck on one sensor

Three places in `renderOpening` draw two symmetric leaves today, all of them from a single `amt` / `tone`:

- **Double-sash swing window** (`sash: "double"`) — the left and right casement leaves. The obvious next case: a window with a contact on each leaf is at least as common as a two-sensor slider.
- **Hinged shutter** (`swingShutter()`, `shutterStyle: "swing"`) — two louvered panels, one per jamb. Note this one would need a *third* entity on the opening, since the shutter already has its own `shutterEntity`; worth deciding whether that is worth the config surface.
- Biparting / converging sliders — **done** in #151.

**A swing door is not on this list**: it is a single hinged leaf, so there is nothing to split. Two-leaf French doors are not modelled at all — that would be a new opening shape rather than an extension of this, and belongs in its own issue if anyone wants it.

## Roughly what it takes

For the double-sash window:

1. Widen the gate — `openingHasTwoPanels(o)` becomes true for a swing window with `sash: "double"`, not only for the two-panel slider styles. Everything downstream keys off this one predicate, including the editor field, so it is the single place to change.
2. The right-hand leaf in `renderOpening`'s double-sash branch reads `amt2` / `tone2` instead of `amt` / `tone` — the same edit the slider branches already have.
3. README row, and tests mirroring the ones #151 added (per-leaf travel, per-leaf accent, the second state ignored where there is only one leaf).

## Naming, while it is cheap

The current names are panel-shaped and a sash is not a panel:

- `openingHasTwoPanels` / `sliderStyleHasTwoPanels`
- editor label **"Second panel"**, helper *"Its own sensor for the other panel"*

Worth moving to something neutral — "leaf" is the word the codebase already uses for both (*"the leaf/sash and swing arc"*) — as part of this, rather than shipping a window field that talks about panels.

The **config key needs no change**: `secondaryEntity` is already generic, which is why none of this is urgent and why there is no migration to regret. That is the reason this was split out rather than bolted onto #151.

## Why not in #151

That PR is scoped to the reporter's door, has a fix outstanding, and someone has been waiting on it since 8 Aug. Growing it to a second element type mid-review would delay a working feature to gain something with no deadline and no lock-in.

@superflyer11 — offering this to you first if you want it, since you built the mechanism and the context is fresh. No obligation.

### LINKED ISSUE #162: Visual indication of offline devices
Hello,

First of all, congratulations on the work you’ve done.

I have a suggestion for improvement: displaying offline devices differently. For example, for a ceiling light, here are two possible ways to indicate that it is offline:

Image

Personally, I don’t have a preference between the two. I just thought of these two options, so I wanted to share the ideas with you so that you can choose one (or potentially allow the configuration to be applied globally to a floor plan).

Thanks!

nicosandller:
Three independent reports, one PR. They touch different files and can be read in any order.

## Per-leaf sensors for the other two-part openings — closes #159

#151 gave a sliding opening a sensor per panel, but the predicate deciding an opening *had* two parts was slider-shaped. It is now `openingHasTwoLeaves`, and a hinged double answers yes — a casement window's two sashes, and the **double door** #168 added, which did not exist when the issue was written. Each leaf draws its own arc, travel and accent, so a window with a contact per sash renders left-open / right-shut.

**Hinged shutters** get the same through a `shutterSecondaryEntity` of their own — the issue flagged this as an open decision, and it needs a *third* entity because `shutterEntity` is taken. Its own key because the shutter is a layer *over* the opening: a double casement behind a pair of shutters has four leaves.

`openingClearFraction` counts both sashes, so one open leaf of a pair clears half the opening for light. Naming moved panel→leaf as the issue asked (`sliderStyleHasTwoLeaves`, `secondLeafOf`, a **Second leaf** field); the config key `secondaryEntity` was already generic, so there is no migration.

## Offline devices look offline — closes #162

An entity reading `unavailable` or `unknown` — or missing from Home Assistant entirely — was drawn **exactly** like a device someone had switched off. New plan-wide `offlineStyle`: `dim` (default), `strike` (dimmed with a diagonal, the reporter's second mock-up), or `none`.

The default changes how an offline device looks on upgrade, deliberately: a dead bulb and a bulb someone turned off were the same picture, and the plan gave that answer confidently. The mark's colour is `--fp-offline-mark`.

## A header that doesn't cost a row — closes #152

`ha-card`'s header is a fixed ~76px whatever the title says, and unreachable from outside its shadow root. `compactHeader: true` declines it: the title becomes a chip in the plan's top-left, and the floor buttons lay out as a row in the same strip. Both then cost zero layout height. Off by default, since the title then sits over the drawing.

## Verification

`npm test` (1025 tests, 45 new), `npm run typecheck`, `npm run build` clean.

Rendering checked in a browser against the built bundle: a casement with one sash open draws one arc and one flat leaf; a shutter with a contact per panel folds one and leaves the other across the glass; `offlineStyle: none` is pixel-identical to today; compact header reclaims the header's height and holds with long titles and floor names.

The dev container grows fixtures for all of it — a shared **Left leaf** / **Right leaf** contact pair driving a casement, a double door and a pair of shutters at once, plus `sensor.flaky_sensor` and a deliberately missing entity for the two flavours of offline.

---

*Multi-reading devices (#180) and the `overlayScale` default (#179) were written after this merged and live in #186, not here.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ screen size.
 <img width="240" height="358" alt="conditionals" src="https://github.com/user-attachments/assets/11d359b6-de8c-483c-8763-105ddf7d915b" />
 
 - **Animated doors & windows** — bind a contact `binary_sensor` or `cover` and openings swing, slide or roll with their real state, partial positions included.
+  - (${\color{red}NEW!}$) **A sensor per leaf** — anything with two leaves takes a second contact and draws them independently: a casement window with one sash open and one shut, a double door ajar on one side, a pair of shutters with one folded back.
+- (${\color{red}NEW!}$) **Offline devices read as offline** — an entity that is unavailable, unknown, or gone from Home Assistant is dimmed (or crossed out), instead of looking exactly like a device someone switched off.
 - (${\color{red}NEW!}$) **Furniture** — 26 gray line-art diagrams (table, sofa, bed, stove, stairs, tv…), each bindable to an entity, in a searchable picker. Every one is a plain JSON file of numbers you can copy: draw your own in the editor's paste box, use it straight away, and open a PR when it's good. No SVG, so nothing you paste can run anything.
 - **Areas** — trace room polygons that color live from an entity, and link them to Home Assistant areas to scope entity pickers and bulk-add devices.
 - **Live position trackers** — map one or two distance sensors (mmWave / radar) onto a marker that moves across the plan in real time.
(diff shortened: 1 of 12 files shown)


nicosandller:
Screenshot 2026-08-18 at 9 14 41 PM


Tested.

PR #210, opened 2026-08-23
Floors: the switcher can live on the stairs (closes #121)

### LINKED ISSUE #121: Make stairs actionable to go to next floor
I see an arrow pointing up or down on the stairs, makes me naturally want to click/tap it to view the next floor.

nicosandller:
The ask was to tap the stairs and see the next floor. Rather than inventing an action type for it, the control that already changes floor becomes something you can put there — your option **B**.

```yaml
floors:
  - { id: ground,   name: Ground floor, icon: mdi:stairs-down, … }
  - { id: upstairs, name: Upstairs,     icon: mdi:stairs-up,   … }
floorSwitcher:
  x: 300           # canvas units — on the staircase you drew
  y: 170
  direction: row
  width: 26
  height: 26
```

`Floor.icon` replaces a button's text with a glyph, which is what lets it be small enough to sit on a drawn staircase. The floor's name becomes the tooltip, so naming it survives the shrink.

**Unplaced, nothing changes.** Without `x`/`y` it is the fixed corner control it has always been — outside the zoom wrapper, pinned to the card. With both, it joins the drawing: canvas units, remapped under `rotation`, travelling with the plan, counter-scaled against zoom-to-room. One coordinate on its own is not a position and falls back to the corner rather than guessing the other axis.

| Key | What it does |
| --- | --- |
| `x`, `y` | Position in canvas units. Both, or neither. |
| `direction` | `column` (default) or `row`. |
| `width`, `height` | Button size in canvas units. Unset, each button sizes to its own label. |
| `hidden` | Draw no switcher at all. |

## Two things I found while building it

**Why B and not A.** Option A needed a custom action in `tap_action` — but that key is edited by HA's own `ui_action` selector, and the action list is a *closed union* in `custom-card-helpers` (`toggle`, `navigate`, `url`, `more-info`, `call-service`, `fire-dom-event`, `none`, `toggle-menu`) with no extension point. A custom `action: "floor-up"` would sit in a control that cannot represent it: blank at best, and plausibly overwritten the first time someone opens that form. Verifying which needs a logged-in HA, so it wasn't something to ship on a guess.

**The switcher is card-level**, while every draggable element in the editor lives in one of the floor's arrays. So it is deliberately *not* a `Sel`: making it one would add a case to `_applyElementPatch`, `_selectionLabel`, the drag snapshot, the delete and copy/paste paths and the selection panel — for a thing that can only ever be moved. It gets its own pointer handling instead, which still snaps to the grid and still commits once so a drag is one undo step.

## In the editor

The switcher is previewed on the canvas and you drag it where you want it — **dashed** while it stands in for the corner, **solid** once placed. Dragging from there is what places it, so there's no separate "enable" step to find. Its shape and the hide toggle are under **Project → Floor switcher** (only shown on a plan with more than one floor); each floor's icon sits beside its short label in that floor's own settings.

Position is deliberately not a form field: it is a place on the drawing, so it is set the way every other placed thing is.

## Verification

`npm test` (1159, 11 new), `npm run typecheck`, `npm run build` clean.

Checked in a browser against the built bundle:

- an unplaced switcher renders in the corner exactly as before;
- a placed one renders as two icon buttons **on the staircase**, active one accented, floor names as tooltips, and clicking the "up" icon switches floor;
- in the editor, the preview starts dashed in the corner, and a real pointer drag places it — config goes from unset to `{ x: 560, y: 180 }`, snapped to the grid, and the preview turns solid.

Branched off `main`; independent of #187.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ screen size.
 <img width="444" height="313" alt="night" src="https://github.com/user-attachments/assets/1590b710-d88f-4a34-986b-b08640a45f4c" />
 
 
-- **Multiple floors** — per-floor elements with a switcher in both the editor and the card.
+- **Multiple floors** — per-floor elements with a switcher in both the editor and the card. (${\color{red}NEW!}$) Drag that switcher onto the plan — onto the staircase you drew, as icon buttons — instead of leaving it in the corner.
 - **Background image** — trace over a floor-plan scan, per floor, with adjustable opacity.
 - (${\color{red}NEW!}$) **Skins** — restyle the whole plan from one line of config: `default` follows your Home Assistant theme, `odnetnin` is chunky charcoal on cream, `pastel` is soft and low-contrast, `tron` is neon on near-black. Colors you set on an element yourself always win.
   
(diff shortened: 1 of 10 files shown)


nicosandller:
Closing this — it solves the wrong problem.

Placing the switcher on the stairs still makes you *pick a floor*. What the issue actually asks for is to *move up or down* with a click, which is a different thing: the control should mean "go up from here", not "show me floor 2".

Replacing it with something much smaller: a config option on the `stairs` furniture that makes it change floor when clicked. No new placement model, no changes to the switcher.

The one piece worth keeping is the dev container fixture — the demo plan was single-floor, so nothing could exercise this at all. That carries over to the new branch.

PR #212, opened 2026-08-24
History replay: scrub the plan through recorder history

### LINKED ISSUE #51: Replay history
As a user, I want to be able to scrolly through and play back the history of devices, so that I can get a visual and holistic understanding of events in my home

TruthOf42:
> **Note:** description written by the maintainer after review — @TruthOf42, please correct anything I've got wrong or rewrite it in your own words.

Adds an optional history replay mode: a timeline and transport controls that scrub the whole plan backwards through Home Assistant's recorder, so you can watch what the house actually did over the last hour rather than only what it's doing now.

## How it works

The design choice that keeps this from touching every drawing function: rather than teaching each renderer about time, replay builds a **synthetic `hass` for timestamp T** and hands it to the existing render path. Everything that already knows how to draw "the house right now" draws "the house at 14:32" with no changes.

- `history-service.ts` fetches `history/history_during_period` for the mapped entities only, and indexes events per entity so `getStateAt(T)` is a lookup rather than a walk over every event.
- `state-provider.ts` swaps `LiveStateProvider` for `HistoryStateProvider` when replay is on.
- `playback-controller.ts` / `replay-controller.ts` own the transport (play, pause, seek, jump, speed).
- `replay-panel.ts` / `history-timeline.ts` are the UI — a scrubbable timeline with per-event markers and an expandable event log.

## Config

Off by default; nothing changes for anyone who doesn't opt in.

```yaml
historyReplay:
  enabled: true          # show the replay controls
  lookbackSeconds: 3600  # how far back the window reaches
  defaultSpeed: 1        # simulated seconds per real second; 1 is real-time
  debug: false           # log replay lifecycle to the console
```

Documented in the README under [History replay](README.md#history-replay).

## Known limitations

- The expanded timeline draws a DOM node per event with no windowing, so a long lookback with a busy plan will get heavy.
- The history parser accepts several response shapes beyond the one `history_during_period` actually returns.
- The card bundle grows ~14% (443 → 503 kB, +13.9 kB gzip) and everyone downloads the replay engine whether or not they enable it. A dynamic import on `historyReplay.enabled` is the escape hatch if it grows further.


amadeobriones:
I run a fork and tested this against `main`: merges clean, type-checks, 1192/1192. The synthetic
`RenderHass` approach is nice — feeding history through the existing render path, and having
`_isOn` / `_openingAmount` delegate to `entityIsActive` / `resolveOpeningAmount` rather than
reimplementing them, means replayed and live state can't drift apart.

One structural thought: `floorplan-card.ts` goes from 2093 to 3464 lines here, about 65% growth in
the largest file in the repo, even though the PR already adds `src/replay-history/`. Three of the
additions don't touch `this` at all — `_normalizeReplayWindow`, `_parseReplayInputValue`,
`_formatReplayInputValue` — so they'd move into that module as-is and be directly unit-testable.
(`_getDefaultReplayWindow` and `_getReplaySpeedForRange` read config off the component, so they'd
need a parameter first.)

Also, `@mdi/font` is added as a devDependency, but its only mention anywhere in the repo is
`package.json` itself — nothing under `src/`, `vite.config.ts`, `vitest.setup.ts` or `docker/`
references it. Possibly a leftover. (For anyone reading the diff nervously: it's a devDependency,
so nothing reaches the shipped bundle.)

Lastly the PR body is empty — for something this size, a couple of lines on what the UI does, or a
screenshot from the dev container, would make it much easier to review.

nicosandller:
Really nice work @TruthOf42 — the shape of this is right, and building it as "hand the renderers a fake `hass` for time T" is what keeps it from touching every drawing function.

Copilot already covered the correctness stuff (the half-wired render paths, the live-state fallback, the 1970 timestamps, the prepare.mjs URL). I checked those and agree — I reproduced the live-state fallback locally, it's real. So I'll only add what it didn't catch.

### Blocker: the test suite doesn't run any more

`vite.config.ts` switches Vitest to jsdom for **every** test, and `jsdom@30` needs Node 22+. On Node 20 the whole suite dies before a single test runs:

```
Test Files  no tests
Errors      14 errors
```

Put main's config back on this branch and it's `1168 passed`. So this isn't the new tests failing — it's all the old ones never starting.

Two things to fix:
- Pin jsdom to a version that still runs on Node 20 (25 or 26).
- Don't switch the whole suite. Put `// @vitest-environment jsdom` at the top of `history-replay.test.ts` and leave everything else alone — it's much faster too.

### The slowness you noticed

I think it's three things stacking up:

1. `getStateAt` walks every event, every frame. Dragging the playhead re-walks the whole list on each frame. Sorting events per entity once and binary-searching would make this basically free.
2. `_buildRenderHass()` builds a fresh states object on every single render — **including when replay is off**. So this currently slows the card down a bit for people who never turn the feature on.
3. `setConfig` clears the history cache and immediately kicks off a new fetch. HA calls `setConfig` on every keystroke in the editor, so typing in the config box fires a history request per character.

(1) and (3) are probably what you were feeling.

### Smaller things

- **`defaultSpeed` doesn't mean what it sounds like.** The speed is `window length / 30`, so the whole window always plays in ~30 seconds. In the demo config `defaultSpeed: 1` is really 240x. @PKeller34 asked for speed relative to real time on a log slider, so it'd be worth making `1` actually mean 1x.
- **`// REVIEW:` notes** are still in the source in four places — look like leftover working notes.
- **Console logging.** Copilot flagged the one line that dumps attributes; more broadly there are ~15 `console.info` calls that fire on every load, seek, play, pause and speed change. Worth dropping most of them before release.
- **Replay colors are hardcoded** (`#f4b400`, `#7b1fa2`, …) instead of going through `skins.ts` like the rest of the card, so they'll look wrong on some skins.
- **The history parser handles a lot of shapes that don't exist** — compact `e`/`s`/`a`/`lu`/`lc` keys, bucket arrays, bare rows. `history/history_during_period` returns one shape. That's a lot of untested code to maintain.
- **The expanded timeline draws a DOM node per event** with no windowing, so a long lookback will get heavy.

### On shipping it as a beta

You asked about getting this out without giving it to everyone — HACS supports this directly. Tag it `v0.8.0-beta.1` and mark it a pre-release on GitHub; only users who tick "show beta versions" in HACS will see it. No fork needed.

One last thing worth not losing: Copilot's *suppressed* list (collapsed by default) has a comment on `floorplan-card.ts:1364` about the `display` default flipping from `badge` to `iconRipple`. That's unrelated to replay and would change existing users' plans, so it should probably come out of this PR.

nicosandller (changes requested):
Really nice work @TruthOf42 — the shape of this is right, and building it as "hand the renderers a fake `hass` for time T" is what keeps it from touching every drawing function.

Copilot already covered the correctness stuff (the half-wired render paths, the live-state fallback, the 1970 timestamps, the prepare.mjs URL). I checked those and agree — I reproduced the live-state fallback locally, it's real. So I'll only add what it didn't catch.

### Blocker: the test suite doesn't run any more

`vite.config.ts` switches Vitest to jsdom for **every** test, and `jsdom@30` needs Node 22+. On Node 20 the whole suite dies before a single test runs:

```
Test Files  no tests
Errors      14 errors
```

Put main's config back on this branch and it's `1168 passed`. So this isn't the new tests failing — it's all the old ones never starting.

Two things to fix:
- Pin jsdom to a version that still runs on Node 20 (25 or 26).
- Don't switch the whole suite. Put `// @vitest-environment jsdom` at the top of `history-replay.test.ts` and leave everything else alone — it's much faster too.

### The slowness you noticed

I think it's three things stacking up:

1. `getStateAt` walks every event, every frame. Dragging the playhead re-walks the whole list on each frame. Sorting events per entity once and binary-searching would make this basically free.
2. `_buildRenderHass()` builds a fresh states object on every single render — **including when replay is off**. So this currently slows the card down a bit for people who never turn the feature on.
3. `setConfig` clears the history cache and immediately kicks off a new fetch. HA calls `setConfig` on every keystroke in the editor, so typing in the config box fires a history request per character.

(1) and (3) are probably what you were feeling.

### Smaller things

- **`defaultSpeed` doesn't mean what it sounds like.** The speed is `window length / 30`, so the whole window always plays in ~30 seconds. In the demo config `defaultSpeed: 1` is really 240x. @PKeller34 asked for speed relative to real time on a log slider, so it'd be worth making `1` actually mean 1x.
- **`// REVIEW:` notes** are still in the source in four places — look like leftover working notes.
- **Console logging.** Copilot flagged the one line that dumps attributes; more broadly there are ~15 `console.info` calls that fire on every load, seek, play, pause and speed change. Worth dropping most of them before release.
- **Replay colors are hardcoded** (`#f4b400`, `#7b1fa2`, …) instead of going through `skins.ts` like the rest of the card, so they'll look wrong on some skins.
- **The history parser handles a lot of shapes that don't exist** — compact `e`/`s`/`a`/`lu`/`lc` keys, bucket arrays, bare rows. `history/history_during_period` returns one shape. That's a lot of untested code to maintain.
- **The expanded timeline draws a DOM node per event** with no windowing, so a long lookback will get heavy.
- we need to add this replay feature as an optional checkbox on "project" configs.


### On shipping it as a beta

You asked about getting this out without giving it to everyone — HACS supports this directly. Tag it `v0.8.0-beta.1` and mark it a pre-release on GitHub; only users who tick "show beta versions" in HACS will see it. No fork needed.

One last thing worth not losing: Copilot's *suppressed* list (collapsed by default) has a comment on `floorplan-card.ts:1364` about the `display` default flipping from `badge` to `iconRipple`. That's unrelated to replay and would change existing users' plans, so it should probably come out of this PR.

nicosandller:
Also please dont forget to add a description and screenshots to help with the review.

TruthOf42:
I decided to go back and rework this a bit. A big thing I did was remove a large majority of the replay-history specific code out. I hope to have an update PR next week. The other big improvements I made was improvements to speed (thanks @nicosandller for pointing out those issues). Its still slower than I like but the history generator is turned up to 11 and I get about 7k events for a 2 hour window. It still works and is pretty usable, just slower than I'd like

nicosandller:
Is this still work in progress @TruthOf42 ?

TruthOf42:
> Is this still work in progress @TruthOf42 ?

Yes

PKeller34:
You guys are **cranking**.   I'll be able to try things out soon.   Right now I'm focused on getting my PoE wall switches commissioned.

TruthOf42:
@nicosandller regarding doing a beta release. My understanding is that we release a particular branch as the beta and then just target that branch for our PRs and such. Do you have any strong feelings about naming conventions for beta branch names and/or version numbers for beta

TruthOf42:
I think I finally have things in a decent place. You'll have to forgive me, my background is mostly web dev (angular) so I'm not familiar with the architecture of home assistant and am leaning on copilot for the majority of my work and then manually refining when I see issues.

nicosandller (changes requested):
Thanks for pushing this out of draft — and for the merge in `fe7e2a4`, which fixes the `floorplan-card.ts` conflict. Verified from a clean merge onto current `main`:

```
npx tsc --noEmit    clean
npm test            1273 passed (16 files)
npm run build       clean
```

### Everything I blocked on last time is fixed

- **The suite runs again.** `jsdom` pinned to `^26.1.0` and the global environment switch replaced with a per-file `// @vitest-environment jsdom` pragma on `history-replay.test.ts`. On Node 20 the whole suite starts and passes — that was the blocker.
- **`_buildRenderHass` no longer runs when replay is off** — it's out of `floorplan-card.ts` entirely (the decoupling commit), so people who never enable this pay nothing per render.
- **`getStateAt` is indexed per entity** (`_eventsByEntity` + `_findLastEventAtOrBefore`) instead of walking every event every frame.
- **`defaultSpeed` is a real-time multiplier now**, with a test pinning it — `1` means 1x, as @PKeller34 asked.
- **The `// REVIEW:` notes are gone** (0 left).
- **The `display` default flip** (`badge` → `iconRipple`) that Copilot had in its suppressed list is out of the diff. Good catch to drop it.
- **It's opt-in via config**, which was the "optional checkbox on project configs" ask — `historyReplay.enabled` defaults to `false`.

The README table for `historyReplay` is exactly what I wanted for the manual-YAML users. Donation links untouched.

---

## Blocking

### 1. The PR has no description and the title is still `Feat/replay history`

This is 22 files and ~4,600 lines — the largest thing in the queue. The body is empty. Everything I know about how this works I've had to reconstruct from the diff and from scattered comments across the thread.

Please write it up before merge: what the feature does, the shape of the design (handing renderers a fake `hass` for time T is genuinely elegant and should be stated), the config surface, and the known limitations. A title that says what it does wouldn't hurt either.

### 2. No screenshots of a UI feature

There's a whole replay panel and timeline here and nothing in the PR shows it. A couple of stills, or better a short GIF of the playhead scrubbing, would make this reviewable now and understandable in a year.

### 3. Seven ungated `console.log` calls on the hot path

`logReplay` (`replay-controller.ts:117`) is a bare `console.log`, and it fires on start, load, **seek**, jump, speed, play and pause. Dragging the playhead logs once per seek — that's a console full of noise for anyone using the feature, and it ships to every user.

One line fixes it: gate the body behind a debug flag, or drop it. The three `console.warn`/`console.error` calls in `history-service.ts` and `replay-controller.ts:349` are fine — those are real failures and should stay.

---

## Non-blocking

<details>
<summary><b>4.</b> <code>setConfig</code> still clears the history cache and restarts on every call</summary>

`floorplan-card.ts:214` does `historyService().clearCache()` then `ensureStarted()`. HA calls `setConfig` on every keystroke in the config box, so a user with replay enabled fires a history query per character typed.

Much less severe than before, because `state.configured` gates the restart and the feature is off by default — so it costs nothing for anyone not using it. But for someone who *has* enabled it, editing the card is still expensive. Worth a guard that skips the clear when the replay-relevant slice of the config hasn't actually changed.
</details>

<details>
<summary><b>5.</b> Five hardcoded colours left outside <code>skins.ts</code></summary>

`replay-panel.ts:176,398` (`#e0e0e0`), `history-timeline.ts:316` (`#03a9f4`), `replay-utils.ts:147,150` (`#ffffff`). Down from nine, so most of this is done — these will still look wrong on some skins.
</details>

<details>
<summary><b>6.</b> Bundle grows ~14% for a feature most users won't enable</summary>

```
main   443.56 kB   gzip 120.55 kB
this   503.10 kB   gzip 134.41 kB
```

+59.5 kB raw, +13.9 kB gzip. The card loads as a single HA resource, so everyone downloads the replay engine whether or not they turn it on. Not a reason to hold the PR — it's a real feature and that's what it costs — but if it grows much further, a dynamic import on `historyReplay.enabled` is the escape hatch.
</details>

Still open from last time and not re-checked, since they're unchanged: the history parser handles several response shapes that `history/history_during_period` never returns, and the expanded timeline draws a DOM node per event with no windowing.

---

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ 33 replay tests, suite green at 1273 |
| 2 | Linked issue / intent | ⚠️ No linked issue — non-blocking, but worth adding for the record |
| 4 | README feature docs | ✅ `historyReplay` table, opt-in and off by default |
| 5 | Screenshots | ❌ See 2 |
| 6 | PR description | ❌ See 1 |

The engineering is in good shape now — items 1 and 2 are the ones I'd most like fixed, and neither is code.

nicosandller (approved):
Approving. The three blockers are cleared — two by me, one that isn't.

## Disclosure: `bcc2aa5` is mine, and so is the PR description

@TruthOf42 — I pushed a commit to your branch and wrote the description that was empty. Both are review feedback you can push back on, not decisions. Please correct anything I've mischaracterised in the write-up.

**The replay log is off by default.** `logReplay` fired on every seek, so dragging the playhead filled the console for every user with no way to quiet it. Now gated behind `historyReplay.debug`, documented in the README, off by default. The `console.warn` / `console.error` calls stay ungated — those are real failures and shouldn't need a config change to surface.

**The history cache survives unrelated edits.** `setConfig` cleared it unconditionally, and HA calls `setConfig` on every keystroke in the config box — so with replay on, editing the card fired a history query per character typed. It now clears only when something a cached window actually depends on has changed: the replay settings, or the set of entities a window is fetched for. The entity set is compared sorted, so rewriting the same entities in a different order isn't a change.

**Six tests**, and I checked they earn their place: reverting the cache guard fails two of them. Five pin the guard (no clear on an unrelated edit or a reordered list; clear on a window change, on disabling replay, and on a new entity), one pins the log gate.

```
npx tsc --noEmit    clean
npm test            1279 passed (16 files)
npm run build       clean
```

## I was wrong about the colours

My last review said five hardcoded colours needed routing through `skins.ts`. Having read them, that was over-broad:

- `replay-panel.ts:176,398` and the `#ddd` in `history-timeline.ts` are `var(--divider-color, #e0e0e0)` — hex *fallbacks inside a `var()`*, which is the pattern used throughout this codebase. Correct as they are.
- `replay-utils.ts:147,150` return `#ffffff` as a value that `history-timeline.ts:43` concatenates with a hex alpha suffix: `box-shadow: 0 0 0 2px ${color}22`. A `var(...)` string there produces invalid CSS. These **can't** be skin tokens without reworking how the glow is built.

The one that was genuinely wrong is fixed: the playhead used `var(--primary-color)` directly and so ignored the skin. It uses `SKIN_ACCENT` now.

## Still outstanding: screenshots

I can't fix this one. Producing them means running the replay UI against a live HA with real recorder data, and I'm not going to fabricate an image of a feature I haven't watched run. It stays on you.

It's the one item on our checklist this PR doesn't meet, and I'm approving anyway rather than holding a feature that is otherwise ready — but please add a couple of stills or a short GIF of the playhead scrubbing before this ships. For a feature whose whole value is motion, it's the difference between someone understanding it from the changelog and not.

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ 39 replay tests, suite green at 1279 |
| 2 | Linked issue / intent | ⚠️ No linked issue — non-blocking, worth adding for the record |
| 4 | README feature docs | ✅ `historyReplay` table, now including `debug` |
| 5 | Screenshots | ❌ Outstanding — see above |
| 6 | PR description | ✅ Written by me; @TruthOf42 to confirm |

Left as-is and worth their own issues rather than blocking: the timeline draws a DOM node per event with no windowing, the history parser accepts response shapes `history_during_period` never returns, and the bundle grows ~14% for a feature most users won't enable — a dynamic import on `historyReplay.enabled` is the escape hatch if it grows further.

Nice piece of work. Handing the renderers a synthetic `hass` for time T instead of teaching every drawing function about time is what made this tractable at all.

TruthOf42:
Screenshots and gif 
[Screencast from 2026-09-01 21-05-59.webm](https://github.com/user-attachments/assets/60e15d08-fb98-4949-8794-6c2b5319eb10)
Screenshot from 2026-09-01 21-04-44
Screenshot from 2026-09-01 21-04-14
Screenshot from 2026-09-01 21-04-04
Screenshot from 2026-09-01 21-03-48
Screenshot from 2026-09-01 21-03-37

TruthOf42:
I think the issue is already linked for this PR. Did you add it? I also added the screenshots and a gif. I was going to create a release branch, and have this PR target that branch and then afterwards we merge the beta back into main. Instead I created this [draft beta release](https://github.com/nicosandller/easy-floorplan/releases/edit/untagged-2af4965afea9c0355116) that targets this branch (feat/replay-history). 

So, two options. 
1) release this PR basically as a beta and then actually merge it when we get more feedback and such
2) continue without beta and just release this as-is and then wait for the feedback to come back from the users

Either way I need a re-approval because I merged main back into this branch due to a merge conflict

nicosandller:
## What testing on the dev container turned up

Running this against a real Home Assistant with two hours of recorder history — rather than reading it — surfaced things the unit tests couldn't. Four rounds of findings and fixes, in `bc51c69`, `a255590` and `a1565e4`, rebased onto @philiplknight's merge of `main`.

```
npx tsc --noEmit    clean
npm test            1313 passed (17 files)
npm run build       clean
```

---

### 1. Playback was unusably slow — and not for the reason it looked like

The obvious suspect was event volume: the demo's tracker sensors change ~11× a minute and are a third of all events. But profiling the real distribution (8,139 events across 20 entities in a 2h window) put the cost somewhere else entirely:

```
getStateAt, 1200 frames            27ms total   (0.02ms/frame — not the problem)
timeline re-render, playhead only  837ms/frame
```

**837ms per frame.** Every marker carried a `passed` class computed from `currentTime`, so moving the playhead invalidated all 8,139 marker templates and lit re-diffed the lot — about 17 seconds of work per second of playback.

Downsampling alone didn't fix it. Halving the events halved the cost and left it ~8× over budget, because the discrete entities (covers, lights, doors ≈ 2,000 events) can't be thinned without losing real transitions.

The fix was to stop rebuilding. `--playhead-pct` on the container, `--marker-pct` on each marker, and `passed` becomes a `clamp()` in CSS that saturates to 0 or 1; the marker list sits behind lit's `guard()`, keyed on the events and window. Moving the playhead now writes one variable.

```
per-frame during playback    837ms -> 0.1ms
```

jsdom doesn't compute CSS, so the passing suite proved nothing about that trick — it was verified in a real browser: a marker behind the playhead resolves to `scale(1.2)`, one ahead to `scale(1)`, and moving only the container variable updates both.

### 2. `numericSteps` — thinning as an option, not a default

Numeric drift can be thinned to N levels of each sensor's own observed range. Discrete states are never touched, and local extrema are always kept so a spike can't flatten into a plateau. This turned out **not** to be what fixed frame rate; it earns its place on first paint:

```
first paint, 2h window   11.5s -> 2.7s at numericSteps: 25
```

Default is unset — lossless. Worth a maintainer decision whether it should ship with a value.

### 3. The lanes were solid white bars

With `numericSteps: 25` already active, `sensor.tracker_distance_x` still drew **932 markers** in a lane that fits about 150. A deadband only removes points that *didn't* move much, and that sensor genuinely swings across its full range several times a minute — so nearly every point is a real, large move and survives.

Thinning the data harder would have coarsened what replay steps through in order to fix how it looks. Instead the timeline now draws at most **150 markers per lane, spent on the largest moves**, leaving the loaded data alone. Discrete lanes are never capped — there's no small change between `on` and `off`, so no honest way to rank them.

### 4. The clock read minutes, so it looked frozen

`dateStyle: "short", timeStyle: "short"` has no seconds:

```
t+ 0s -> 9/2/26, 8:02 AM
t+30s -> 9/2/26, 8:02 AM
t+59s -> 9/2/26, 8:03 AM
```

Near real-time speed it sat on one string long enough to read as "replay isn't running". It shows seconds now, in tabular figures and primary text colour. It was also in the wrong place — while a replay runs the timeline is the only part of the card the eye is on, so **the current time now rides the playhead**.

### 5. The flaky sensor stayed solid — a bug in the thinning I'd just added

`isNumericSeries` bailed on the first non-numeric value. `unavailable` appears 5th in that series, so the whole sensor was classified as *discrete* and exempted from both the deadband and the lane budget. It drew all 435 of its points.

That's backwards: a sensor that drops out is still numeric, and it's also the one most likely to be noisy — the lane that most needs thinning was the only one escaping it. Any battery or Zigbee sensor behaves this way, so it's the common case, not an edge one.

Outages are now ignored when classifying and preserved when thinning. Every drop-out and recovery is kept at any resolution, the deadband doesn't measure drift *across* a gap, and the reading that ends an outage is always kept.

```
sensor.flaky_sensor   435 raw -> 246 at numericSteps:25 -> 150 drawn, 23/23 outages kept
```

**A correction worth recording:** an earlier measurement of this lane reported "150 drawn" and was wrong. The SQL behind it filtered `unavailable` rows out of the sample — precisely the rows that triggered the bug. It looked like verification and wasn't.

### 6. Lanes can be switched off

Clicking a lane's label removes it from the lane view **and from the collapsed summary bar** — a summary still counting a lane you'd switched off isn't a summary of what you're looking at.

The label is a real button — border on hover, pointer cursor, and a dot that fills when the lane is on and hollows when off — because a bare text label that happens to be clickable isn't discoverable. A switched-off lane keeps its row, struck through, so it can be switched back, and a "N lanes hidden / Show all" control appears while anything is hidden so collapsing the lanes can't strand one.

It's a per-session view preference held in the component, not config or storage — it resets on reload. If it should persist, that's a maintainer call between `localStorage` and a config key.

### 7. Smaller

The **Show replay history** button was a right-aligned pill; it's now on the left, shaped like the floor buttons and carrying the same skin tokens, labelled "Replay" with the full description kept as an `aria-label`.

---

### Still open for a maintainer decision

- Whether `numericSteps` should default to a value rather than unset.
- Whether the `clamp()` trick for the passed-marker effect is too clever for this codebase — the simpler alternative is a "played region" overlay instead of per-marker scaling, which changes the look.
- Whether hidden lanes should persist across reloads.
- **Screenshots are still outstanding** — the one checklist item this PR doesn't meet. It needs someone who can watch the replay UI run; I won't fabricate an image of it.

<sub>Prepared with Claude Code.</sub>

nicosandller (approved):

nicosandller:
@TruthOf42 I think we should just release it as is, since there's no UI element to add `replay` at the moment. People won't find it easily so we can leave it as a beta for expert tinkerers.

What do you think?

PKeller34:
I'm all for releasing the alpha as a hidden mode of an update.   I just put together an example of my P&ID idea to post on the other thread discussing the idea.   Animation really spiced it up.   History replay will really add to the flavor and enhance understanding the use case.

This brought a new question up.   What would be a good way to "export" and "import" the history of the entities that show up in a floorplan so I could, for example, provide the floorplan card along with some history data that would allow one to see the card in operation without having to snapshot ALL of my HA environment AND the recorder history?    I guess I could record a screenshot video....

PR #214, opened 2026-08-24
Keep the wall neon outside the doorway mask

### LINKED ISSUE #203: Neon Glow Ignores Wall Openings
Image

The glow effect on the Tron theme follows walls but doesn't respect wall openings or cause the door/window outlines to glow. Ideally the  door and window outlines would glow, but at a minimum it shouldn't continue to glow across empty space as pictured.

amadeobriones:
Addresses #203.

On a skin that sets `--fp-skin-wall-filter`, the wall neon carried straight through doors and
windows, so a plan read as one unbroken glowing outline instead of a room with openings in it.

### Cause

CSS applies `filter` before `mask`. The filter sat on the wall element itself, inside the group that
punches the doorway holes:

```
<g mask="url(#fp-wall-mask-0)">
  <line class="wall fp-wall" ... />   <-- filter here
</g>
```

So each wall was rendered *with its full glow*, and only then was the opening cut out of the result.
The cut clears `WALL_THICKNESS + 4` — 12 units, so ±6 from the centreline — while
`drop-shadow(0 0 4px …)` reaches roughly ±8.5. The couple of units of halo on each side that the mask
never touches ran through every opening.

On a Tron render with a wall at x=150 and a door across it, the mask rect and the wall geometry read
straight off the DOM: the mask clears `x ∈ [144, 156]` and the 5px wall occupies
`x ∈ [147.5, 152.5]`, so the wall body *is* cut correctly. The halo figure is derived rather than
measured — a blur radius of 4 is a Gaussian of sigma 2, which fades out around 6 units, giving
roughly `x ∈ [141.5, 158.5]`. The luminance numbers below are what actually pins it down.

### Measured

Peak luminance sampled inside an opening, against a 7.8 background, on the same plan and skin:

| | peak | above background |
|---|---|---|
| solid wall (reference) | 211.0 | 203.2 |
| before | **35.6** | **27.8** |
| after | **7.8** | **0.0** |

About 17% of full wall brightness — plainly visible against a near-black ground. Skins without a
wall filter never showed it at all, since there is no halo to leak.

### Fix

The filter moves to `.fp-wall-neon`, a group that wraps the masked group, so the hole is punched
first and the glow is computed from the shape that survives.

`editor.ts` has the same defect in a sharper form — `mask` and the filter on the *same* element — at
both the committed-walls site and the wall-being-drawn site, so both are wrapped too. The editor is
meant to preview the glow the plan will actually have, and would otherwise disagree with the card.

Both were checked in a running Home Assistant, on a 37-wall plan with the Tron skin. In the editor
the numbers come out as intended: 37 wall lines, all 37 carrying the doorway mask, all 37 wrapped in
the neon group, computed `filter` on the wall itself `none`, and `drop-shadow(rgb(34, 211, 238) 0px
0px 4px)` on the group above it. Doorways read as real gaps on the canvas.

Only skins that set a wall filter were ever affected. Default and the others set `none` and render
identically before and after.

### What this does not do

#203 also asks that door and window outlines themselves glow. **This PR does not do that** — opening
symbols render outside the filtered group and are untouched here. Left deliberately, as it is a
design decision about how openings should look rather than a rendering defect. That is why this says
*Addresses* rather than *Closes*.

### Testing

`wall-neon-mask.test.ts` pins the structure rather than the pixels: the wall carries no filter of its
own, `.fp-wall-neon` carries it, the mask is nested inside that group, every masked wall in the
editor is wrapped, and the reason is written beside the rule so it does not get moved back. All seven
guards fail against the unfixed source — checked by reverting the fix and re-running, so they cannot
quietly pass forever.

`npx tsc --noEmit` clean. **1175 tests pass** (1168 before, plus the 7 added here).

Both sides are exercised against a real render of the built bundle: the card through the images and
luminance numbers above, the editor in a live Home Assistant as described.

### Screenshots

Same room, same three openings (one window, two doors), same skin per row. Top row Tron before/after;
bottom row the Default skin before/after, which is pixel-identical either way.

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -3267,10 +3267,10 @@ export class FloorplanCardEditor extends LitElement {
               }
               ${
                 this._draft
-                  ? svg`<line x1=${this._draft.x1} y1=${this._draft.y1}
+                  ? svg`<g class="fp-wall-neon"><line x1=${this._draft.x1} y1=${this._draft.y1}
                               x2=${this._draft.x2} y2=${this._draft.y2}
                               class="wall draft" mask=${`url(#${this._wallMaskId})`}
-                              stroke-width=${WALL_THICKNESS} />`
+                              stroke-width=${WALL_THICKNESS} /></g>`
                   : nothing
               }
               ${this._renderAreaDraft()}
@@ -3810,10 +3810,10 @@ export class FloorplanCardEditor extends LitElement {
         <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
               class="wall-hit"
               @pointerdown=${(e: PointerEvent) => this._startDrag(e, { kind: "wall", id: w.id })} />
-        <line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
+        <g class="fp-wall-neon"><line x1=${w.x1} y1=${w.y1} x2=${w.x2} y2=${w.y2}
               class="wall ${selected ? "selected" : ""}"
               mask=${`url(#${this._wallMaskId})`}
-              style=${wallStrokeStyle(w.thickness)} stroke-linecap="round" />
+              style=${wallStrokeStyle(w.thickness)} stroke-linecap="round" /></g>
         ${
           selected
             ? svg`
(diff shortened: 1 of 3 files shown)


amadeobriones:
issue203-before-after

nicosandller (approved):

Issue #143, opened 2026-08-08
Light casting should treat open doors like openings


amitkeret:
First of all, let me say how impressed I am with the quality and features offered by this card. Well done 🥇 

I especially like the recent release which introduced cast lighting.
At the moment, though, doors act as walls and stop the light pool, regardless of open/close status.

Opening - expected light pool behavior:
<img width="281" height="372" alt="Image" src="https://github.com/user-attachments/assets/6bf54334-128b-43fc-9181-bb153a192e2c" />

Open door - wrong light pool behavior:
<img width="272" height="316" alt="Image" src="https://github.com/user-attachments/assets/2dd4ec52-1075-4b77-9ab0-d7996d764161" />

Can I suggest to treat open doors the same as you treat a simple opening in a wall?
Thank you!

nicosandller:
@amitkeret  I'm really happy you like the recent features :). 


I was so excited to build them too. And a lot of these are community driven ideas! If you have any cool ideas yourself make sure to share them and or even open a PR yourself!



Thank you for the feedback, you're absolutely right!  I will work on this improvement soon!


🚀

nicosandller:
Released on `v1.2.8`!

<img width="242" height="269" alt="Image" src="https://github.com/user-attachments/assets/8c420070-4d5e-4842-bd93-f5c2d8caf902" />

PR #215, opened 2026-08-25
Windows let a lamp's glow through, shut or open

diegocjorge:
Hey! Noticed that a lamp's glow (issue #143's wall-cutting) treats a window the same way it treats a door: shut, it blocks the light, and only counts as "open" when its sensor says so. That felt off for a window, since even a closed one is still glass — the light should still spill through, the same way it already does through an open door.

This PR makes glass let a lamp's own glow through it, open or shut, reusing the same `glazed` field sunlight already reads (`openingSunFraction`) rather than introducing a second name for the same physical fact:

- A window is glass by default, same as it already is to the sun — so it lets a lamp's glow through it regardless of its sash. A patio door marked `glazed: true` now does too. A window explicitly marked `glazed: false` (a glass-brick panel, a hatch) stays opaque to a lamp's light exactly as it already was to the sun.
- One exception sunlight doesn't need: a roll-motion window — a blind, shutter, shade, curtain or awning auto-mapped from its `device_class` — is a covering standing in for the glass, not the glass itself, so it keeps blocking a lamp's glow according to its own position instead of reading as always-clear.
- A shutter still wins if it's rolled down, same priority sunlight already gives it: fully closed, it blocks the light regardless of the glass behind it. This isn't window-only — any opening with a `shutterEntity` bound now reads it, so a door with a shutter rolled down blocks a lamp's pool through it too, where it used to pass while the door itself was open.

Before:
SCR-20260825-prpq

After:
SCR-20260825-psos

### DIFF
--- a/src/floorplan-card.ts
+++ b/src/floorplan-card.ts
@@ -75,6 +75,7 @@ import {
   renderSunDimMask,
   wallsLightPassesThrough,
   openingClearFraction,
+  glowClearFraction,
   polygonCentroid,
   trackerSensorReading,
   entityIsActive,
@@ -784,8 +785,16 @@ export class FloorplanCard extends LitElement {
       ? wallsLightPassesThrough(active.walls, active.openings, (o) =>
           // Both leaves, and the travel each style actually has (issue #145):
           // asking `entity` alone left a door whose *second* panel was open
-          // still blocking light outright.
-          openingClearFraction(o, this._openingAmount(o), this._openingSecond(o)?.amount)
+          // still blocking light outright. Glass admits it whole regardless
+          // of sash — a closed window is not a hole, but light still gets
+          // through it. A shutter rolled down overrides that, same as it
+          // does for sunlight.
+          glowClearFraction(
+            o,
+            this._openingAmount(o),
+            this._openingSecond(o)?.amount,
+            o.shutterEntity ? shutterAmount(this.hass?.states[o.shutterEntity], o.shutterInvert) : undefined
+          )
         )
       : active.walls;
     // Lit rooms hold back the night (issue #113): without this the flat dim
--- a/src/render.test.ts
+++ b/src/render.test.ts
@@ -135,6 +135,7 @@ import {
   glowReach,
   wallsLightPassesThrough,
   openingClearFraction,
+  glowClearFraction,
   renderGlowMask,
   renderOpening,
   renderGlow,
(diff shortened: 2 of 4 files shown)


src/render.ts
```diff
@@ -642,6 +642,51 @@ export function openingClearFraction(o: Opening, amount: number, secondAmount?:
...
+ * pinned to fully clear.
+ */
+export function glowClearFraction(
+  o: Opening,
+  amount: number,
+  secondAmount?: number,
+  shutter?: number,
+): number {
+  if (shutter !== undefined && shutter <= 0) return 0;
+  if (openingIsWindow(o)) return 1;
```
nicosandller: A patio door marked `glazed: true` is a wall of glass, and this blocks a lamp's pool through it while letting it through the window beside it.

I get the reasoning in the docstring — `glazed` grew up as sunlight's field. But the word means "is glass", and someone who set it on their French door has already told the card the thing this rule needs to know. Two names for one physical fact is what will make this hard to explain later.

src/render.test.ts
```diff
@@ -4388,6 +4389,55 @@ describe("openingClearFraction (#145 / #143)", () => {
...
+  });
+
+  it("ignores `glazed` — that field is sunlight's, not a lamp's", () => {
+    // A French door is glass to the sun (openingSunFraction), but it is
+    // still a door to a lamp's pool: shut, it blocks like any other door.
+    expect(glowClearFraction({ type: "door", glazed: true } as Opening, 0)).toBe(0);
+    expect(glowClearFraction({ type: "door", glazed: true } as Opening, 0.4)).toBeCloseTo(0.4);
+    // A window stays glass to a lamp even marked `glazed: false` — that flag
+    // only ever meant "block the sun", never "block a lamp's own light".
+    expect(glowClearFraction({ type: "window", glazed: false } as Opening, 0.4)).toBe(1);
```
nicosandller: This is the case I'd push back on hardest. `glazed: false` on a window is documented as "an opaque window like a glass-brick panel or a hatch" — so after this line, a shut hatch blocks the sun but leaks a lamp's pool.

If `glazed` is genuinely sunlight-only, the README line saying so needs to explain why an opaque window still passes lamp light. I think the simpler answer is that it isn't sunlight-only.

src/floorplan-card.ts
```diff
@@ -784,8 +785,16 @@ export class FloorplanCard extends LitElement {
...
-          openingClearFraction(o, this._openingAmount(o), this._openingSecond(o)?.amount)
+          // still blocking light outright. Glass admits it whole regardless
+          // of sash — a closed window is not a hole, but light still gets
+          // through it. A shutter rolled down overrides that, same as it
+          // does for sunlight.
+          glowClearFraction(
+            o,
+            this._openingAmount(o),
+            this._openingSecond(o)?.amount,
+            o.shutterEntity ? shutterAmount(this.hass?.states[o.shutterEntity], o.shutterInvert) : undefined
```
nicosandller, line 796: Not a problem, just worth saying out loud: the shutter check applies to every opening, not only windows. A door with a shutter bound that is rolled down now blocks the pool where it used to pass it while the door was open.

That's the right behaviour, but it's a change existing configs will see and the description only mentions windows. Worth a line there.

nicosandller (changes requested):
Thanks for this — the core idea is right, and the before/after shots make the case well. A shut window is still glass, and a lamp's pool should go through it.

One thing to sort out before it lands: **what counts as glass**. Right now the card can give opposite answers about the same opening depending on which light is asking. I ran the three cases on your branch, shut:

| opening | sunlight | lamp glow |
| --- | --- | --- |
| `type: window` | passes | passes |
| `type: door, glazed: true` (patio door) | passes | **blocks** |
| `type: window, glazed: false` (hatch, glass brick) | blocks | **passes** |

Both of the bottom rows are hard to explain to someone using the card. A glass patio door blocking a lamp while the window next to it doesn't, and an opaque hatch stopping the sun but leaking lamp light, are the same field being read two ways.

I'd suggest one shared "is this opening transparent" rule that both `openingSunFraction` and `glowClearFraction` call, rather than each having its own opinion. That would also close the roll-up hole Copilot found in one place instead of two — worth knowing that hole is **older than this PR**: on `main` today, `openingSunFraction({ type: "window", motion: "roll" }, 0)` already returns 1, so a rolled-down shutter has been passing sunlight all along. Your change doesn't cause it, it just gives it a second home.

Nothing here is a big rewrite — the tests you've written are good and mostly stay as they are once the rule moves. Happy to pair on it if you'd like.

src/render.ts
```diff
@@ -642,6 +642,51 @@ export function openingClearFraction(o: Opening, amount: number, secondAmount?:
...
+  return openingClearFraction(o, amount, secondAmount);
```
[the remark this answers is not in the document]
diegocjorge: Done

src/render.test.ts
```diff
@@ -4388,6 +4389,55 @@ describe("openingClearFraction (#145 / #143)", () => {
   });
 });
 
+describe("glowClearFraction — glass admits a lamp's pool shut or open", () => {
+  it("passes an opaque opening straight through to openingClearFraction", () => {
+    expect(glowClearFraction({ type: "door" } as Opening, 0.25)).toBeCloseTo(0.25);
+    expect(glowClearFraction({ type: "door" } as Opening, 1)).toBe(1);
+    expect(glowClearFraction({ type: "door" } as Opening, 0)).toBe(0);
+  });
```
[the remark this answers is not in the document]
diegocjorge: Done

src/render.ts
diegocjorge: Ok, reused `openingIsGlazed` instead of a separate check. Only exception is the roll-up case above, which needed its own carve-out either way.

src/render.test.ts
diegocjorge: Fixed in the same commit. A `glazed: false` window now stays opaque to a lamp's glow too.

src/floorplan-card.ts
diegocjorge: Updated the PR description to mention that.

diegocjorge:
To test it, I created a floor plan with every window/door type in each of these states:

1. Open
2. Closed
3. Open, with the shutter closed
4. Closed, with the shutter open

Test results:
image

nicosandller (approved):
Thanks for the changes @diegocjorge .


Screenshots and tests on the dev container really helped!

PR #221, opened 2026-08-28
Climate, weather and media_player icons never reflected their state

diegocjorge:
entityIsActive only ever recognised the literal state "on" (or a small per-domain set for lock/vacuum/camera) as active. A climate entity's state *is* its HVAC mode — "cool", "heat", "dry" — never "on", so a running unit always read as off: no active highlight, and a configured iconAnimation never played however it was actually running. A media player reporting "paused"/"idle"/"buffering" instead of "playing" had the same problem — still switched on, read as off.

climate and media_player each get their own active-state set now, mirroring lock/vacuum/camera: every HVAC mode but "off", and every media_player state but "off".

Separately, climate and weather had no state-aware icon at all — DOMAIN_STATE_ICONS only ever expressed a two-way on/off choice, and neither domain's meaning is binary. A climate item always drew the bare thermostat glyph regardless of mode, and a weather item — with no device class and no domain entry — drew entityDefaultIcon's plain fallback circle whatever the sky was doing.

Adds CLIMATE_MODE_ICONS (one icon per HVAC mode, HA's own climate.HVACMode enum) and WEATHER_CONDITION_ICONS (one per condition, HA's own ATTR_CONDITION_* set), and threads the entity's raw state into entityDefaultIcon so it can pick from them. A climate entity in fan_only always spins regardless of iconAnimation — its own fan is what's running, the same physical fact that makes a `fan` domain entity spin by default, just decided from this entity's state rather than its domain; `none` still wins over it.

### DIFF
--- a/docker/config/floorplan-demo.yaml
+++ b/docker/config/floorplan-demo.yaml
@@ -13,7 +13,9 @@
 #                      light.kitchen_lights, light.office_rgbw_lights,
 #                      cover.living_room_window, cover.hall_window,
 #                      cover.garage_door, fan.living_room_fan,
-#                      device_tracker.demo_anne_therese
+#                      device_tracker.demo_anne_therese, climate.hvac,
+#                      weather.demo_weather_south, weather.demo_weather_north,
+#                      media_player.walkman
 #   from configuration.yaml
 #                      sensor.living_area_temperature,
 #                      sensor.living_area_humidity,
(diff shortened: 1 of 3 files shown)


nicosandller (approved):
**Approve** — clean, well-scoped fix.

Behaviour is unchanged for every domain and state outside the three this touches: each new branch in `entityDefaultIcon` returns early only for `climate` / `weather`, the `ACTIVE_STATES` additions sit beside the existing lock/vacuum/camera sets, and the `resolveIconAnimation` line is gated below the `none` and fail-closed guards. The icon and active-state maps track HA's own `HVACMode` and `ATTR_CONDITION_*` enums; unrecognised values fall back to the previous glyph rather than to nothing; card and editor stay in sync because both resolve through `resolveItemIcon`, which now threads `st.state`. Every branch has a test, `tsc` is clean, and the full suite is green (1202).

Rendered every state against the `main` and PR bundles — before/after for climate, weather and media_player here: https://claude.ai/code/artifact/d910c002-55bc-4d6f-a51e-ffe1cc5da870

Two small notes, neither blocking:

- `entityDefaultIcon` for `media_player` still keys only on/off, so `paused` / `idle` now draw `mdi:television-play`. "It's on" is the right semantic and it matches the amber highlight, but a distinct pause glyph would be a nice follow-up.
- `climate` counts `heat` / `cool` / … as active regardless of `hvac_action`, so a thermostat in `heat` mode but currently idle (setpoint satisfied) reads as active. That matches HA's own Tile card and is overridable with a state rule on `hvac_action`, so it's fine — just flagging the choice.

### One request before merge

The demo plan (`docker/config/floorplan-demo.yaml`) has no `climate.*`, `weather.*` or `media_player.*` device on either floor, so none of this is visible in the dev container — only in the unit tests. The `demo` integration already ships `climate.heatpump` / `climate.hvac` / `climate.ecobee`, a `weather.*` entity and several `media_player.*`.

Could we add a few of these to the default plan — and ideally have the sample-data automation walk a climate entity through its HVAC modes the way the covers already cycle position — so the feature has a visible home, the same way the per-leaf sensors and multi-reading badges do?

nicosandller:
@diegocjorge I approved but could you add these new feature examples on the dev container so we can keep an aye on them in the future?

diegocjorge:
Added the pause icon (`mdi:television-pause`) and also gave `idle` its own icon (`mdi:television`). Everything else that counts as active (`on`, `playing`, `buffering`) keeps the plain play icon (`mdi:television-play`).

On the `hvac_action` point, I think that one really depends on the kind of HVAC setup. Where I live it's common to have an individual A/C unit per room, and for those, being in heat/cool mode does mean the unit is on, whether or not it's actively heating/cooling at that instant. So I'd rather leave the current behavior as the default (it also matches HA's own Tile card) and let anyone who wants idle-vs-running to show differently add a state rule on `hvac_action`, like you mentioned.

nicosandller (approved):

PR #231, opened 2026-08-31
Drag: focus the canvas without scrolling it out from under the pointer

### LINKED ISSUE #229: Editor: drag picks the wrong element (or the Area underneath) when the canvas is scrolled
## What happens

In the editor, dragging an item barely moves it and then stops. Very often the
room **Area** underneath gets selected and dragged instead of the item I grabbed.

It only happens when the canvas is scrolled inside its container — a tall floor
in a short window. Scroll the canvas back to the top and dragging is fine.

## Why

`_onCanvasDown`, `_onOverlayDown` and `_startDrag` call `this._canvasWrap?.focus()`
*before* resolving the pointer:

```ts
this._canvasWrap?.focus();
const raw = this._toVirtual(ev, false);      // or _resolvePick(ev, sel)
```

`focus()` scrolls the element into view, and `_toVirtual` maps `ev.clientX/Y`
through the SVG's **live** `getScreenCTM()`. So layout moves between the event
being dispatched and the CTM being read, and the resolved point is off by the
scroll delta. On my plan a 91px scroll became a 100.8-unit error — far enough
that the point misses the item's hit box and lands in the Area polygon below it,
which is exactly the "the room moved instead" symptom. The subsequent drag then
computes deltas from that wrong origin.

Five call sites do this (editor.ts 410, 1112, 1322, 1503, 2998).

## Fix

`focus({ preventScroll: true })` — focus without the scroll, so layout is stable
across the pick. PR follows.

## Version

easy-floorplan 1.5.5 via HACS, Home Assistant 2026.8, Chrome on macOS. The drag
code is byte-identical in 1.5.3/1.5.4/1.5.5, so this is not a recent regression.

sasha-id:
Fixes #229.

`focus()` scrolls its element into view. Every pointerdown path calls it before
resolving the pointer, and `_toVirtual` maps `clientX/Y` through the SVG's live
`getScreenCTM()` — so when the canvas is scrolled inside its container, layout
shifts between the event being dispatched and the CTM being read, and the
resolved point is off by the scroll delta.

Measured on a tall floor in a short window: 91px of scroll became a 100.8-unit
error. That is enough to miss the dragged item's hit box and land in the Area
polygon beneath it, which is why the room drags instead of the device, and why
the gesture then appears to "barely move" — the deltas are computed from the
wrong origin.

`{ preventScroll: true }` keeps the focus behaviour and drops the scroll. Five
call sites, all of which want focus but not a scroll: the three pointerdown
paths, the Escape handler that parks focus back on the canvas, and the floor
`<select>` that hands focus back after switching.

## Testing

`npm run typecheck` and `npm test` (1221 tests) pass. Verified by A/B on a real
install: with the canvas scrolled, stock picks the Area under the item and the
patched build picks the item.

I did not add an automated test — the repo's suite is pure-unit with no DOM
environment, and this behaviour needs a scrolled layout plus a live
`getScreenCTM()`. Happy to add one if you'd like the DOM harness introduced.

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -407,7 +407,7 @@ export class FloorplanCardEditor extends LitElement {
     if (!isTypingPath(ev.composedPath())) return;
     ev.preventDefault();
     ev.stopPropagation();
-    this._canvasWrap?.focus();
+    this._canvasWrap?.focus({ preventScroll: true });
   };
   private _onFocusIn = (ev: FocusEvent) => {
     // While the fullscreen popover is up, anything that pulls focus outside the
@@ -1109,7 +1109,7 @@ export class FloorplanCardEditor extends LitElement {
     if (ev.button !== 0) return;
     // One gesture at a time: a second touch must not hijack the state machine.
     if (this._gesturePointer !== null) return;
-    this._canvasWrap?.focus();
+    this._canvasWrap?.focus({ preventScroll: true });
     const raw = this._toVirtual(ev, false);
 
     if (this._tool === "wall") {
@@ -1319,7 +1319,7 @@ export class FloorplanCardEditor extends LitElement {
     if (this._tool !== "select") return;
     ev.stopPropagation();
     if (this._gesturePointer !== null) return;
-    this._canvasWrap?.focus();
+    this._canvasWrap?.focus({ preventScroll: true });
     // Endpoint/vertex handles always operate on that single element; every
     // other click goes through the overlap-aware picker (issue #52).
     const explicitHandle = endpoint != null || areaVertex != null;
@@ -1500,7 +1500,7 @@ export class FloorplanCardEditor extends LitElement {
     // preventDefault suppresses native mousedown focusing, so focus explicitly.
     ev.preventDefault();
     if (this._gesturePointer !== null) return;
-    this._canvasWrap?.focus();
+    this._canvasWrap?.focus({ preventScroll: true });
     const pick = this._resolvePick(ev, sel);
     this._selectForPointer(ev, pick);
     this._drag = {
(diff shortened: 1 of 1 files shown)


nicosandller (approved):
Approving. Nice diagnosis — the write-up on *why* the point drifts (layout shift between event dispatch and the `getScreenCTM()` read) is the useful part, and the 91px → 100.8-unit measurement makes the failure concrete.

I merged this onto `main` alongside #232: clean auto-merge, `tsc --noEmit` clean, `npm test` 1229/1229.

**Verified:**

- **Complete.** All five `_canvasWrap?.focus()` sites are covered — `git grep '_canvasWrap?.focus()'` returns nothing on this branch.
- **Safe degradation.** `preventScroll` is a hint; engines that ignore it fall back to exactly today's behaviour, so there's no regression risk on older webviews.
- **Right fix for the cause.** The three pointerdown paths are the ones that actually matter (they focus and then immediately call `_toVirtual`); the Escape handler and the floor `<select>` don't have the ordering problem but shouldn't be scrolling the canvas either. Consistent to change all five.

**Checklist**

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ⚠️ None added — accepted, see below |
| 2 | Linked issue / intent | ✅ Fixes #229, and does what it asks |
| 3 | Donation links untouched | ✅ README not touched |
| 4 | README feature docs | ✅ N/A — bug fix, no user-facing feature |
| 5 | Screenshots | ⚠️ None; A/B described in prose. Non-blocking |
| 6 | PR description | ✅ Clear, with a measurement |

On tests: your reasoning is right and I agree — the suite is pure-unit with no DOM environment, and this needs a scrolled layout plus a live `getScreenCTM()`. Not worth introducing a DOM harness for a five-token change. Thanks for flagging it rather than leaving it silent.

One heads-up, not a change request here: #232 re-introduces this same class of bug at the *other* end of the gesture, by deferring a `PointerEvent` to a later frame and re-deriving its point against a CTM that may have changed in between. I've left the detail on that PR. Nothing for you to do in this one — merge order doesn't matter.

PR #232, opened 2026-08-31
Drag: the host renders once per gesture, not once per pointermove

### LINKED ISSUE #230: Editor: drag freezes and jumps on heavy configs — one config-changed per pointermove
## What happens

On a heavy config, dragging an element in the editor doesn't track the cursor.
It starts fine, then freezes and jumps: the element sits still for ~100ms while
the cursor moves on, then snaps to catch up, repeatedly, for the whole gesture.

Frame-tracking a screen recording of one 14-second session: the dragged element
changed position on 26 frames out of 832, in clusters separated by stretches
where it did not move at all.

## Why

`_applyDrag` → `_emitFloor` → `_emit` fires `config-changed` on **every**
`pointermove`. HA's edit dialog answers each one by re-rendering the preview and
echoing the config back into `setConfig`, which `structuredClone`s every floor.

The part that makes it severe: **the preview re-renders whatever card is being
edited, and that can be much more than this card.** Mine is one card inside a
20-card vertical stack, so a single `config-changed` re-renders the floorplan
*plus* 19 other cards. Each emit stalls the main thread far longer than a frame.

Two independent costs, both on the pointermove path:

1. **How often the host is asked to re-render** — once per pointermove.
2. **How often `_applyDrag` runs** — also once per raw pointermove. When the
   pointer outruns the renderer this builds an input backlog that grows for the
   length of the gesture, so the lag accumulates rather than settling.

The editor's own render is not the problem: I measured `performUpdate` at 0.4ms
(p95 2.7ms) at 1930x1152 with `devicePixelRatio` 2.

## Measurements

Harness: the editor mounted beside a live preview card, wired like HA's dialog
(async echo, preview `setConfig` on every `config-changed`), Chrome, retina
viewport 1930x1152. To model the stack the preview's `setConfig` was given an
80ms cost. 3-second drag:

| | host re-renders during the drag | p95 frame | worst frame | frames > 33ms |
| stock (emit per pointermove) | 19+ | 83ms | 99ms | 22 / 149 |
| emit deferred to pointerup | 1 | 18ms | 47ms | 6 / 177 |

Under a 6x CPU throttle, a 50-move drag at 125Hz: 2312ms of main-thread time and
50 preview re-renders with an 18px trail, versus 865ms and 1px once the per-move
emit is gone. Coalescing the moves into one per animation frame is worth a
further ~20% (2825ms → 2258ms at 6x; 3501ms → 3008ms at 12x) and bounds the
backlog so the lag can't accumulate.

## Suggested fix

During a drag, keep the config local and emit once on `pointerup`, and coalesce
`_applyDrag` into one call per animation frame. `_config` is a reactive `@state`,
so the editor's own canvas still repaints every frame — only the host's preview
waits for release. Trade-off: no live preview mid-drag.

There's a side benefit: the echo currently lands mid-gesture and clears the undo
history, so undo after a drag doesn't work. Deferring the emit fixes that too.

PR follows.

## Version

easy-floorplan 1.5.5 via HACS, Home Assistant 2026.8, Chrome on macOS.

sasha-id:
Fixes #230.

Dragging an element emits `config-changed` on every `pointermove`. The host answers each one by re-rendering the card being edited and echoing the config back through `setConfig`, which deep-clones every floor. Where that render is cheap the cost hides. Where it isn't — my card is one child of a `vertical-stack`, so a single emit re-renders all 19 of its siblings — each emit stalls the main thread for ~80–100ms, and the drag freezes and jumps rather than following the pointer.

Two changes, one per layer.

**`FrameCoalescer` (new, `src/frame-coalescer.ts`)** applies at most one drag move per animation frame. A pointer reports faster than a browser renders — a 1000Hz mouse against a 60Hz display is 16 events per frame — so handling every raw event doesn't make the drag smoother, it makes it *later*: the queue grows for as long as the gesture lasts. Coalescing bounds it at one event, so the drag can fall at most one frame behind however fast the pointer moves. The scheduler is injected so it can be tested without a DOM.

What it carries is the move already resolved to canvas coordinates (`DragMove {x, y, altKey}`), not the `PointerEvent`. Resolving needs `getScreenCTM()`, which is live — scrolling the canvas (a plain wheel does exactly that) or zooming changes it — so transforming a frame later would read the *new* matrix against the *old* pointer position and place the element off by the scroll delta.

Both pointerup paths call `settle()` **while `_drag` is still set**, so the element lands on the last position the pointer actually reported rather than on whatever the previous frame caught. `_cancelGesture` drops the queued frame — a canceled drag stays a no-op — and `disconnectedCallback` drops it, flushes, and ends the gesture, since HA's dialog reparents the editor and that can land mid-drag.

**`_emitFloor`** keeps the config local during a drag and emits once on release (`_flushDrag`). The canvas still follows the pointer every frame: `_config` is `@state`, so assigning it repaints the editor without involving the host at all.

### Measurements

A harness mounts the editor next to a preview card and wires them the way HA's edit dialog does (config echo, preview re-set on every `config-changed`), with an 80ms `setConfig` standing in for a heavy host. 3s drag, ~200Hz input, retina 1930×1152, a 72-item floor:

| | host re-renders during the drag | p95 frame | median frame | frames over 33ms |
| before | 36 | 85.3ms | 83.5ms | 35 / 48 |
| after | 1 | 25.8ms | 22.4ms | 6 / 130 |

Medians of three runs per build.

Release is unchanged: one emit, and the element lands where the pointer left it (0px error, no snap-back when the host echo arrives).

Simulating the host is the part that matters. The editor's own update costs ~0.4ms at that resolution — it was never the bottleneck, and a harness whose preview is a single cheap card shows almost no difference between these two builds.

### Trade-off, deliberate

A preview outside the editor no longer updates mid-drag; it catches up on release. The editor canvas is authoritative during the gesture, so nothing is lost where the user is looking.

### Notes

- The coalescer carries a resolved point rather than the `PointerEvent`, so nothing it holds depends on state that can change before delivery. An earlier revision of this PR held the event and got this wrong; see the review thread for the measured failure.
- `setConfig`'s existing echo-detection needs no change. It's the emit that stops during a drag, so there is no echo to guard against.
- 10 unit tests for `FrameCoalescer` (manual scheduler: newest-of-burst, one frame per burst, re-queue, settle-without-waiting, settle consumes the pending frame, cancel, and a queued `null` delivered rather than read as "nothing queued"). The editor wiring itself isn't covered — the suite has no DOM environment — so it was verified in the harness above and by running the built bundle on a live HA install.
- `npm run typecheck` clean, `npm test` 1231 passed.

Independent of #231; either can go first.

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -231,6 +231,29 @@ interface Drag {
   snapshot?: FloorplanCardConfig;
   /** The redo stack as it stood before the drag's history push cleared it. */
   priorFuture?: FloorplanCardConfig[];
+  /**
+   * Set if anything emitted while this drag was live. A drag normally emits
+   * nothing until it is released, which is what lets cancel restore the
+   * pre-drag config locally; if something did emit, the host has moved on and
+   * has to be told to go back.
+   */
+  emitted?: boolean;
+}
+
+/**
+ * One drag move, already resolved to canvas coordinates.
+ *
+ * Resolved where the event arrives rather than where it is applied: a pointer
+ * event carries screen coordinates, and turning those into canvas coordinates
+ * needs the SVG's CTM, which is live — scrolling the canvas (a plain wheel
+ * does exactly that) or zooming changes it. Transforming a frame later would
+ * read the *new* CTM against the *old* pointer position and place the element
+ * off by the scroll delta.
+ */
+interface DragMove {
+  x: number;
+  y: number;
+  altKey: boolean;
 }
 
 type Marquee = Rect;
@@ -379,8 +402,8 @@ export class FloorplanCardEditor extends LitElement {
    * frame cannot finish, so the element falls further behind the longer the
    * drag lasts.
    */
-  private _dragMoves = new FrameCoalescer<PointerEvent>(rafScheduler, (ev) => {
-    if (this._drag) this._applyDrag(ev);
+  private _dragMoves = new FrameCoalescer<DragMove>(rafScheduler, (move) => {
+    if (this._drag) this._applyDrag(move);
   });
   /** A drag changed the config and the host has not been told yet. */
   private _dragDirty = false;
@@ -443,6 +466,20 @@ export class FloorplanCardEditor extends LitElement {
     this.removeEventListener("keydown", this._onHostKeyDown);
     window.removeEventListener("focusin", this._onFocusIn);
     if (this._applyResetTimer !== null) clearTimeout(this._applyResetTimer);
+    // HA's dialog reparents the editor, so this can land mid-drag. Removal
+    // takes the pointer capture with it, so the gesture is over either way:
+    // hand the host what the drag accumulated (_config is otherwise its only
+    // copy of the edit, and a save sees only what config-changed delivered),
+    // then end the gesture. Dropping the queued move rather than settling it
+    // commits the position the canvas last actually drew — settling would be
+    // safe too, now that a queued move carries no DOM dependency, but it would
+    // commit a position that was never rendered. Clearing _gesturePointer is
+    // not optional: it gates every new gesture, and the pointer it belongs to
+    // can no longer report.
+    this._dragMoves.cancel();
+    this._flushDrag();
+    this._drag = null;
+    this._gesturePointer = null;
     this._resetPinch();
     super.disconnectedCallback();
   }
@@ -836,6 +873,7 @@ export class FloorplanCardEditor extends LitElement {
   private _lastEmitted?: FloorplanCardConfig;
 
   private _emit(config: FloorplanCardConfig): void {
+    if (this._drag) this._drag.emitted = true;
     this._config = config;
     // Recompute here, not just in setConfig: real HA deep-equal-skips the
     // setConfig echo of our own emission, so entities bound during the
@@ -1212,7 +1250,16 @@ export class FloorplanCardEditor extends LitElement {
     this._drag = null;
     if (drag?.moved && drag.snapshot) {
       this._history = this._history.filter((c) => c !== drag.snapshot);
-      this._emit(drag.snapshot);
+      if (drag.emitted) {
+        this._emit(drag.snapshot);
+      } else {
+        // Nothing emitted while the drag was live, so the host still holds
+        // exactly this config: restore locally and spare it a full re-render.
+        // _lastEmitted stays as it is — it still describes the last thing
+        // actually emitted.
+        this._config = drag.snapshot;
+        this._watchedEntities = collectWatchedEntities(drag.snapshot);
+      }
       // The push at first movement cleared the redo stack; a canceled drag
       // must be a complete no-op, so put it back.
       this._future = drag.priorFuture ?? [];
@@ -1263,7 +1310,7 @@ export class FloorplanCardEditor extends LitElement {
       this._marquee = { ...this._marquee, x1: raw.x, y1: raw.y };
       return;
     }
-    if (this._drag) this._dragMoves.push(ev);
+    if (this._drag) this._dragMoves.push({ ...this._toVirtual(ev, false), altKey: ev.altKey });
   }
 
   private _onCanvasUp(ev: PointerEvent): void {
@@ -1413,9 +1460,8 @@ export class FloorplanCardEditor extends LitElement {
     return m;
   }
 
-  private _applyDrag(ev: PointerEvent): void {
+  private _applyDrag(p: DragMove): void {
     const drag = this._drag!;
-    const p = this._toVirtual(ev, false);
     // First *effective* movement: snapshot for undo now, not at pointerdown,
     // so a plain selection click — including the ~1px jitter real clicks and
     // taps produce — doesn't spam history or wipe the redo stack. Threshold
@@ -1433,7 +1479,7 @@ export class FloorplanCardEditor extends LitElement {
     // corners of other walls travel along (Alt detaches), so dragging a room
     // corner stretches the room (issue #30).
     if (drag.endpoint) {
-      const attach = ev.altKey ? [] : (drag.attached ?? []);
+      const attach = p.altKey ? [] : (drag.attached ?? []);
       // The moving cluster must not be a snap candidate for itself — the
       // dragged corner would stick to its own last emitted position.
       const moving = new Set<string>([
@@ -1509,7 +1555,7 @@ export class FloorplanCardEditor extends LitElement {
     // Whole-wall drag: shared corners of neighboring walls follow (issue #30),
     // unless Alt detaches or the neighbor is itself part of the selection
     // (then it already translated with the group).
-    if (drag.attached?.length && !ev.altKey) {
+    if (drag.attached?.length && !p.altKey) {
       const walls = (patch.walls ?? f.walls).map((w) => {
         let out = w;
         for (const a of drag.attached!) {
@@ -1557,7 +1603,7 @@ export class FloorplanCardEditor extends LitElement {
       this._cancelGesture();
       return;
     }
-    if (this._drag) this._dragMoves.push(ev);
+    if (this._drag) this._dragMoves.push({ ...this._toVirtual(ev, false), altKey: ev.altKey });
   }
 
   private _onOverlayUp(ev: PointerEvent): void {
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -143,6 +143,7 @@ import {
   type Sel,
   type SelKind,
 } from "./editor-geometry";
+import { FrameCoalescer, rafScheduler } from "./frame-coalescer";
 import { applyCardConfig } from "./editor-save";
 import type { AreaEntityScope } from "./editor-forms";
 import {
@@ -373,6 +374,17 @@ export class FloorplanCardEditor extends LitElement {
 
   private _drag: Drag | null = null;
   /**
+   * Drag moves are applied once per animation frame. A pointer reports faster
+   * than the browser renders, and applying every raw move only queues work the
+   * frame cannot finish, so the element falls further behind the longer the
+   * drag lasts.
+   */
+  private _dragMoves = new FrameCoalescer<PointerEvent>(rafScheduler, (ev) => {
+    if (this._drag) this._applyDrag(ev);
+  });
+  /** A drag changed the config and the host has not been told yet. */
+  private _dragDirty = false;
+  /**
    * Where the last plain selection click landed, for click-cycling through
    * overlapping elements (issue #52). Cleared whenever the pointer lands
    * somewhere else, so cycling only happens on repeat clicks in one spot.
@@ -506,7 +518,26 @@ export class FloorplanCardEditor extends LitElement {
 
   /** Live change to the active floor's elements (no history snapshot — for dragging). */
   private _emitFloor(partial: Partial<Floor>): void {
-    this._emit({ ...this._config, floors: this._patchFloors(partial) });
+    const config = { ...this._config, floors: this._patchFloors(partial) };
+    // A drag owns the config until pointerup. Emitting per move asks HA to
+    // re-render whatever card is being edited — for a stack, every child of it
+    // — and to echo the config back through setConfig, which stalls the gesture
+    // far longer than a frame. The canvas still follows the pointer: _config is
+    // @state, so assigning it repaints the editor without involving the host.
+    if (this._drag) {
+      this._config = config;
+      this._watchedEntities = collectWatchedEntities(config);
+      this._dragDirty = true;
+      return;
+    }
+    this._emit(config);
+  }
+
+  /** Hand the host what a drag accumulated — once, on release. */
+  private _flushDrag(): void {
+    if (!this._dragDirty) return;
+    this._dragDirty = false;
+    this._emit(this._config);
   }
 
   private _patchFloors(partial: Partial<Floor>): Floor[] {
@@ -1168,6 +1199,9 @@ export class FloorplanCardEditor extends LitElement {
    * between — is dropped, so a canceled drag leaves no trace in undo.
    */
   private _cancelGesture(): void {
+    // Whatever the drag accumulated is about to be replaced by its snapshot.
+    this._dragMoves.cancel();
+    this._dragDirty = false;
     this._gesturePointer = null;
     this._draft = null;
     this._draftTracker = null;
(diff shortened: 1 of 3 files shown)


nicosandller (changes requested):
Requesting changes — two issues, both small to fix. The approach is right and the measurements are convincing; I want the gesture-lifecycle edges closed before this merges.

Merged onto `main` with #231: clean auto-merge, `tsc --noEmit` clean, `npm test` 1229/1229. The 8 `FrameCoalescer` tests are well chosen — the manual scheduler is the right call, and `settle()`-while-`_drag`-is-still-set is exactly the subtle part to have gotten right.

---

## Blocking

### 1. Holding the `PointerEvent` across a frame is not safe — the CTM is live

From your Notes:

> The coalescer holds a `PointerEvent` across a frame. That's safe here: `_toVirtual` reads only `clientX/Y` against the live SVG CTM, and `_applyDrag` additionally reads `altKey` — all persisted on the event, none of them live.

The event *fields* are indeed static. The problem is the other operand: `_toVirtual` transforms those static `clientX/Y` through `getScreenCTM()` read **at delivery time**, a frame later. If the CTM changes between `push()` and delivery, the point resolves wrong — which is the same failure #229/#231 describes, just moved to the release end of the gesture.

Two CTM-changing paths are reachable with the button held:

- **Plain wheel.** `_onCanvasWheel` returns early without `preventDefault()` unless Ctrl/Cmd is held, so the wheel natively scrolls `.canvas-wrap` (`overflow: auto`).
- **Ctrl/Cmd+wheel** → `_setZoom`.

Scenario: grab an item, wheel-scroll to reach a distant part of the plan, release. `settle()` transforms the stale `clientX/Y` through the *new* CTM, so the element lands off by the scroll delta — and because `_flushDrag()` emits right after, that wrong position is the one persisted. Before this PR, `_applyDrag` ran synchronously inside the `pointermove` dispatch, where the CTM still matched the event.

**Fix:** resolve at `push()` time and coalesce the result, not the event:

```ts
private _dragMoves = new FrameCoalescer<{ x: number; y: number; altKey: boolean }>(
  rafScheduler,
  (m) => { if (this._drag) this._applyDrag(m); },
);
// call site: this._dragMoves.push({ ...this._toVirtual(ev, false), altKey: ev.altKey });
```

`_applyDrag` takes the resolved point instead of calling `_toVirtual` itself. This also drops the `PointerEvent` retention question entirely, and makes the coalescer's payload a plain value.

### 2. `disconnectedCallback` neither cancels the queued frame nor flushes the drag

`disconnectedCallback` does its other teardown (listeners, `_applyResetTimer`, `_resetPinch`) but leaves both new pieces of state alone. Your own `_resetPinch` comment establishes that this is a real path, not a theoretical one:

> HA's dialog reparents the editor, which fires disconnected/connected without a second `firstUpdated`

Two failure modes if that lands mid-drag:

- **Teleport.** The queued rAF still fires with `_drag` set. The detached SVG's `getScreenCTM()` returns null, `_toVirtual` returns `{ x: 0, y: 0 }`, and `_applyDrag` computes a delta from the origin — a large bogus jump applied to the whole selection.
- **Silent data loss.** `_dragDirty` means `_config` holds the *only* copy of the user's edit. `applyCardConfig` (`src/editor-save.ts`) saves the dialog's `_cardConfig`, which is fed solely by `config-changed` — so an unflushed drag is dropped on save with no error. This window didn't exist before, because every move had already been emitted.

**Fix:** two lines in `disconnectedCallback`:

```ts
this._dragMoves.cancel();
this._flushDrag();
```

(Order matters: drop the queued frame first, then hand the host what's accumulated.)

---

## Non-blocking

<details>
<summary><b>3.</b> <code>_cancelGesture</code> now emits a config the host already has</summary>

`this._emit(drag.snapshot)` was necessary when every move was emitted. Since this PR emits nothing during a drag, the host still holds exactly `drag.snapshot`, so Escape-cancelling (or the two-finger cancel from `_onWrapPointerDown`) fires a `config-changed` carrying no change — which triggers precisely the whole-`vertical-stack` re-render this PR exists to avoid, on a path where nothing happened.

`_emit` also sets `_config`, `_watchedEntities` and `_lastEmitted`, so it can't just be deleted. Restoring locally is enough:

```ts
this._config = drag.snapshot;
this._watchedEntities = collectWatchedEntities(drag.snapshot);
```

Leave `_lastEmitted` alone — it still correctly describes the last thing actually emitted. Nothing else can emit mid-drag (`_handleKeyDown`'s `gestureActive` guard blocks keyboard mutations, `_emitFloor` is guarded), so the host cannot have drifted.
</details>

<details>
<summary><b>4.</b> <code>null</code> doubles as the "nothing queued" sentinel in <code>FrameCoalescer</code></summary>

`push(null)` on a nullable `T` is accepted, stored, then silently dropped by `if (value !== null)` in both `push`'s callback and `settle`. Can't bite with `PointerEvent`, but the class is exported as a general utility with an injected scheduler, so a future nullable `T` would hit it. A separate `private _hasPending = false` removes the ambiguity. (Finding 1's payload change makes this even easier to get right.)
</details>

<details>
<summary><b>5.</b> <code>pending</code> is unused and reports handle state, not value state</summary>

Nothing in `src/` reads it — only `frame-coalescer.test.ts`. It also answers a slightly different question than its doc comment (`True while a value is waiting for its frame`): it returns `this._handle !== null`, so it describes whether a frame is scheduled. Either drop it or define it as `this._pending !== null`.
</details>

---

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ 8 focused unit tests for the new unit; editor wiring uncovered (no DOM env) and disclosed |
| 2 | Linked issue / intent | ✅ Fixes #230, and addresses the actual cause |
| 3 | Donation links untouched | ✅ README not touched |
| 4 | README feature docs | ✅ N/A — perf fix, no user-facing feature |
| 5 | Screenshots | ✅ N/A — the before/after table is more useful here |
| 6 | PR description | ✅ Excellent, including the deliberate trade-off |

On the trade-off you flagged — a preview outside the editor no longer updating mid-drag: keep it as-is, no config flag. The editor canvas is authoritative where the user is looking, and a flag would mean maintaining both paths forever to preserve the slower one.

Happy to merge once 1 and 2 are addressed; 3–5 at your discretion.

sasha-id:
Thanks — both blocking findings are real, and finding 1 is worse than my Notes claimed. That bullet was wrong: I checked that the event's *fields* are static and stopped there, without noticing the CTM is the other operand and is read a frame later.

Reproduced it before fixing it. Harness at 2.5× zoom so `.canvas-wrap` overflows, drag an item, leave one `pointermove` queued, scroll the wrap 150px (what a plain wheel does), release. 150px is 278 canvas units at that zoom, and the last 5px of real pointer travel is 12 units:

| | applied Δy | applied Δx (no h-scroll) |
| --- | --- | --- |
| before this PR | 12 | 12 |
| this PR as reviewed | **288** | 12 |
| with the fix | 12 | 12 |

The element landed the full scroll delta away, and `_flushDrag()` persisted exactly that. Δx staying correct isolates it to the CTM.

Fixed as you suggested — `_toVirtual` runs at `push()` time, the coalescer carries `DragMove {x, y, altKey}`, and `_applyDrag` takes the resolved point. Those were its only three uses of the event.

I expected resolving per raw move rather than once per frame to cost something, so I re-measured instead of assuming. It doesn't — three runs of each build, same 3s drag at ~200Hz over a 72-item floor with an 80ms host `setConfig`:

| | host re-renders | p95 frame | median frame | frames over 33ms |
| before this PR | 36 | 85.3ms | 83.5ms | 35 / 48 |
| coalescing the event (as reviewed) | 1 | 25.7ms | 22.2ms | 6 / 132 |
| coalescing the resolved point | 1 | 25.8ms | 22.4ms | 6 / 130 |

Medians of three; the two coalescing builds are inside each other's run-to-run spread. (These supersede the single-run figures in the description, which I've updated — the 19.4ms p95 there was an optimistic sample, not a baseline.)

**Finding 2** fixed as prescribed. One thing I checked first: whether a `config-changed` dispatched from a *detached* element reaches HA at all — if it didn't, the flush would clear `_dragDirty` into the void and make the data loss permanent rather than fixing it. It does reach: `hui-element-editor.ts` binds the listener directly on the editor element (`configElement.addEventListener("config-changed", …)`), so the target is always in the event path. That's what makes `disconnectedCallback` the right place for it rather than `connectedCallback`, so I've recorded it in the comment there.

Two things came out of implementing it that go beyond the two lines, both because your finding 2 and your finding 3 interact:

**The flush has to end the gesture too.** `_flushDrag()` at disconnect emits mid-drag, which is exactly the case finding 3's local restore assumes cannot happen. Applied as just the two lines, the sequence is: the reparent flushes the moved position to the host → `_drag` is still live, and a reparent is precisely when pointerup goes missing (your own comment on the `buttons === 0` guards says so) → so the recovery path is `_cancelGesture`, which now restores the pre-drag config locally and emits nothing → the canvas shows the old position while the dialog holds the half-finished one, with the snapshot already filtered out of history so undo can't reach it either.

I built that intermediate rather than argue it. Drag an element +72,+72, reparent the editor mid-gesture, then deliver the buttons-less move that stands in for the lost pointerup:

| | after the reparent | after the cancel |
| --- | --- | --- |
| this PR as it stands | editor 772,852 · host never told · gesture still live | editor 700,780 · host 700,780 — agree, but the drag is gone |
| findings 2 + 3, literally | editor 772,852 · host 772,852 | editor **700,780** · host **772,852** — diverged |
| with the fix | editor 772,852 · host 772,852 · gesture ended | editor 772,852 · host 772,852 |

So `disconnectedCallback` also clears `_drag` and `_gesturePointer`. That is not tidiness: removal takes the pointer capture with it, so the gesture really is over, and `_gesturePointer` gates every new gesture (`_onCanvasDown`) while the pointer it names can no longer report — clearing `_drag` without it would lock the editor out of gestures entirely, since the `buttons === 0` guard would then have nothing left to fire on. The edit now survives the reparent, which is what the flush was for.

**`_cancelGesture` decides rather than assumes.** Rather than leave the local restore resting on a whole-file argument that the next change can silently break — which is how the above happened — `Drag` gained an `emitted` flag that `_emit` sets while a drag is live. Cancel restores locally when nothing emitted (your finding 3, the normal path) and emits the snapshot when something did. It also closes a pre-existing crack I found while checking every call site: `_patchConfigLive` calls `_emit` directly with no drag guard, so a second finger on the properties panel mid-drag could already desynchronise a later cancel.

I did consider gating the restore on `configsEqual(drag.snapshot, this._lastEmitted)` instead, since the helper exists. It doesn't work: `_emit` deletes the empty legacy arrays, while `_config` always carries them as `[]`, so that comparison is false even in the normal case and the restore would always fall back to emitting.

Related, and your call: now that the payload is resolved at push time it carries no CTM dependency, so `settle()` at disconnect would also be safe. I kept your `cancel()` — it commits the position the canvas last actually drew, where settling would commit one that was never rendered — and rewrote the comment there, since the reason it originally gave (the detached-tree CTM) stopped being true the moment finding 1 was fixed.

**3, 4 and 5** all applied.

- 3: restores `_config` and `_watchedEntities` locally, `_lastEmitted` untouched as you said.
- 4/5: a `_hasPending` flag replaces the `null` sentinel and now backs `pending`, so it reports value state rather than handle state — which is what its doc always claimed. Two tests added for a queued `null` (through a frame, and through `settle`). `push`/`settle`/`cancel` share `_cancelFrame`/`_deliverPending` helpers so the three paths can't drift.

Trade-off stays as-is, no flag — agreed, and for the reason you gave.

`tsc --noEmit` clean, `npm test` 1231 passed.


Pushed as `3177e8c`, and the description is updated to match. Ready for another look whenever you have time — and I can split the `emitted` flag into its own commit if you would rather review it separately from the disconnect handling.

nicosandller (approved):
Approving. All five findings are addressed, and two of them are solved better than I proposed.

Verified from a clean merge onto `main`:

```
npx tsc --noEmit    clean
npm test            1231 passed (15 files)
npm run build       clean
files changed       src/editor.ts, src/frame-coalescer.ts, src/frame-coalescer.test.ts — nothing else
```

**1. CTM resolved at arrival (blocking).** Fixed at the root: `_applyDrag` now takes a `DragMove`, and both push sites (`_onCanvasMove`, `_onOverlayMove`) resolve through `_toVirtual` at push time. `grep` confirms no `PointerEvent` survives a frame anywhere in the drag path, so the wheel-scroll-mid-drag scenario has nothing stale left to transform. The `DragMove` doc comment records *why* the point is resolved where it arrives, which is the part that keeps this from regressing.

**2. `disconnectedCallback` (blocking).** Fixed, and you went further than the two lines I suggested — also clearing `_drag` and `_gesturePointer`. You're right that those are needed: leaving `_gesturePointer` set would gate every subsequent gesture on a pointer that can no longer report. Dropping the queued move rather than settling it is also the better call, and for the reason you give — it commits the position the canvas last actually drew.

**3. `_cancelGesture` (non-blocking) — your solution is better than mine.** I suggested restoring `_config` unconditionally, resting on the claim that nothing can emit mid-drag. The `drag.emitted` flag doesn't rest on that claim at all: it observes what actually happened, so if some future path does emit during a drag, cancel still corrects the host instead of silently leaving it ahead. That removes an invariant I was asking you to maintain by hand. Take the credit for this one.

**4 & 5. `FrameCoalescer` (non-blocking).** `_hasPending` disambiguates a queued `null`, `pending` now reports value state to match its doc comment, and `_cancelFrame` / `_deliverPending` remove the duplication between `push` and `settle`. The two new tests pin the exact ambiguity — a queued `null` delivered on tick, and settled rather than swallowed — which is the right pair.

### Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ 10 unit tests; editor wiring still uncovered (no DOM env) and still disclosed |
| 2 | Linked issue / intent | ✅ Fixes #230, at the cause |
| 3 | Donation links untouched | ✅ README not touched |
| 4 | README feature docs | ✅ N/A — perf fix, no user-facing feature |
| 5 | Screenshots | ✅ N/A — the before/after measurements table is more useful |
| 6 | PR description | ✅ Updated for this revision, including the note that an earlier revision held the event and why that was wrong |

The one thing still uncovered is the editor wiring itself — the CTM fix and the `disconnectedCallback` teardown are both verified by reading and by your harness, not by the suite, because there's no DOM environment. That's a pre-existing gap in the project rather than something to hold this PR for, and you disclosed it both times. Worth its own issue if we ever want editor gesture tests.

Good work, and thanks for the care on the write-up — the measurement table and the explicit trade-off section made this much faster to review.

sasha-id:
On the coverage gap you noted: filed as #233 rather than leaving it implicit here.

Worth flagging one thing from it, because it inverts the obvious fix — I probed both DOM shims before proposing anything. jsdom has no `getScreenCTM` at all, and happy-dom has one that always returns the identity matrix, with zero layout in both. So a regression test for this bug written under happy-dom would have run, passed, and told us nothing: a CTM that never changes and a canvas that can never scroll cannot express it. Real-browser tests (vitest browser mode) would catch it; a DOM shim would just make untested code look tested.

Not asking you to take that on as part of this PR — it is a separate call, and the issue says so.

Issue #244, opened 2026-09-02
Feature Request: Directed Occupancy Ripple


jjj120:
I have a presence sensor and a motion sensor which are both occupancy devices in HomeAssistant.

I like the ripple animation with the normal occupancy devices. But my sensors are directed, they only see about 170°, not the full 360°. 
It would be nice to be able to set the direction and width of the ripple animation to not always have the ripple all the way around the sensor icon.
In my case, the sensor is mounted to the wall, therefore the ripples go through the wall, even tough the sensor cannot see there.

nicosandller:
@jjj120 thanks for the feedback. I have a rough idea of what you mean but can you share a few drawings of how you imagine it ? What controls do you imagine for the setup? Maybe this is an opportunity to make the animation even cooler? Any ideas ?

jjj120:
I already added a pull request (#247) with a screenshot and an implementation of how I imagined it to look

PR #247, opened 2026-09-02
Feature Request: Directed Occupancy Ripple

jjj120:
Solves #244

I added the option to limit the size of the ripple animations. There are two more configuration parameters when enabling the ripple effect:

image
<br/>
image

These can set/limit the angular width of the ripple effect. This is intended for motion or presence sensors mounted on a wall. 

image


nicosandller (changes requested):
Thanks for this — and for opening #244 first with the reasoning. "My sensor is on a wall and the ripple goes through it" is a good, concrete problem, and a masked arc is the right shape of answer. The screenshots showing the editor fields and the result make it easy to see what you mean.

Verified from a clean merge onto `main`: `tsc` clean, `npm test` 1376 pass, `npm run build` clean, merges cleanly.

Three things before it can go in.

---

## Blocking

### 1. The two new values are written into a `style` attribute unsanitized

`render.ts:4849`:

```ts
style="…--fp-ripple-color:${cssColorOr(color, SKIN_ACCENT)};--fp-ripple-direction:${direction};--fp-ripple-width:${width}"
```

The line directly above it is the convention this needs:

```ts
const size = overlayLength(cssNumber(sizePx, DEFAULT_RIPPLE_SIZE), scale);
```

`sizePx` is passed through `cssNumber`, and `color` through `cssColorOr` — `direction` and `width` go in raw. TypeScript says they are `number`, but these come from YAML, and the caller only guards with `??`, which catches `null`/`undefined` and nothing else. So a string reaches the attribute intact. I checked rather than assumed:

```
rippleDirection: "0; background:url(https://evil.example/x.png)"

style="width:80px;height:80px;--fp-ripple-color:#fff;
       --fp-ripple-direction:0; background:url(https://evil.example/x.png);
       --fp-ripple-width:360"
```

That is a live declaration, and that particular one makes an outbound request when the card renders — an IP leak from any shared plan. It also breaks an invariant this repo states and tests explicitly, in `css-safe.adversarial.test.ts`:

> `cssColor — INVARIANT: nothing accepted can break out of a style declaration`

The fix is small — run both through `cssNumber` with their defaults, the way `sizePx` already is. Clamping is worth considering while you are there: `width` outside `0…360` and `direction` outside `0…360` are not meaningful, and clamping at the sink is what `itemLabelSize` and `rippleSize` do.

### 2. No tests for the feature

The only change in `render.test.ts` is updating two existing calls to the new signature. Nothing asserts that a direction or width reaches the mask, that the defaults are 0/360, or that a hostile value is neutralised — so the whole feature and the fix above are uncovered.

The comment on the test you edited is worth reading, because it is about exactly this:

> `// path: dropping the argument at one of the badge/item call sites would otherwise still pass.`

Suggested minimum: direction and width appear as the custom properties for a normal config; the defaults apply when the keys are absent; and a string like the one above comes out sanitized.

### 3. The README does not mention either key

Two new config options with no entry in the item table and no prose. Anyone editing YAML by hand has no way to discover them, and the section on ripples still describes a ring that always goes all the way round.

---

## Notes, not blocking

<details>
<summary><b>4.</b> Every existing ripple now goes through a mask, including full-circle ones</summary>

The `conic-gradient` mask is applied unconditionally, so a plan that has never set these keys renders its ripples through a 360° mask it did not have before. At `width: 360` that should be fully opaque and visually identical — but it adds a compositing layer to every ripple on every plan, and a full-turn conic gradient is exactly where a seam artifact shows up if it is going to.

Worth an eye on a real plan with a default ripple before and after, and worth considering skipping the mask entirely when `width >= 360`.
</details>

<details>
<summary><b>5.</b> On Copilot's note about the exported signature — I disagree, but there is a simpler option</summary>

Copilot flagged inserting `direction`/`width` before `rings`/`scale` as a breaking API change. In practice it is not: the card ships as a single bundle to Home Assistant, `renderRipple` has no external consumers compiled against it, and every call site is in this repo. No overloads needed.

That said, `rings` and `scale` already have defaults — putting the two new parameters *after* them, also with defaults, would have left all four call sites untouched. Your call; I would not hold the PR for it.
</details>

<details>
<summary><b>6.</b> Copilot's two small ones are right</summary>

`editor-forms.ts:1081` — "Ripple Width" is title case where its neighbours are sentence case ("Ripple size", "Ripple direction").

`types.ts:667` — the two new JSDoc lines have no trailing period, unlike `rippleSize` above them. While you are there, *"Direction of the center of the ripple if its width is not 360° in degrees"* reads as though 360 is the unit; something like "Direction the arc is centred on, in degrees clockwise from the top. Default `0`." would be clearer.
</details>

<details>
<summary><b>7.</b> CI has not run on the current head</summary>

`cafe4cd` got checks; `9027ed2`, `8c6c322` and `544e32a` have none, so the head is unverified by CI. That matters a bit more now that `main` has a `browser-tests` job. A push (or re-running the workflow) should sort it — flagging so it is not mistaken for a green run.
</details>

---

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ❌ None for the new behaviour — see 2 |
| 2 | Linked issue / intent | ✅ Solves #244, and matches what was asked for |
| 4 | README feature docs | ❌ Neither key documented — see 3 |
| 5 | Screenshots | ✅ Editor fields and the rendered arc |
| 6 | PR description | ✅ Brief and clear, links the issue |

The idea and the implementation shape are both good — a mask on the ring is neater than redrawing the arcs. Sort 1–3 and I will take another look. Happy to push the `cssNumber` change myself if you would rather I did, but it is a two-line fix and it is your PR.

nicosandller (changes requested):
Good turnaround — all three blockers are addressed, and I verified each rather than reading the commit messages.

```
npx tsc --noEmit    clean
npm test            1377 passed (17 files)
npm run build       clean
```

**The injection is closed.** `cssNumber` plus the clamp, and I re-ran the same probe from my last review across a wider set:

```
"0; background:url(https://evil.example/x.png)"  -> --fp-ripple-direction:0
'0" onload="alert(1)'                            -> --fp-ripple-direction:0
"0/*"  /  "abc"  /  null  /  undefined           -> defaults
-720 / 1e6                                       -> clamped to 0 / 360
170 / 90                                         -> passes through unchanged
```

Nothing escapes the declaration, and legitimate values are untouched. Clamping at the sink as well was the right call.

Tests, the README entry, the sentence-case label and the JSDoc periods are all in.

---

## Two small things, then I'll approve

I wrote both and tried to push them to this branch to save you the round trip, but my sandbox blocked the push — so here they are to apply. Neither is a criticism of the fix; they're the parts that keep it from quietly coming undone later.

### 1. A test for the sanitizing itself

The new tests cover threading the values through and the clamp, but not a hostile string — which is the case `cssNumber` was added for. The clamp alone does **not** cover it: `Math.min(Math.max("0; background:red", 0), 360)` is `NaN`, not the default. So `cssNumber` could be dropped in a refactor and every assertion currently in the file would still pass. I checked that, and it does.

```ts
  // Both values land in a `style` attribute, so they are exactly the kind of
  // config the adversarial suite exists for (css-safe.adversarial.test.ts):
  // "nothing accepted can break out of a style declaration".
  it("cannot be used to inject a style declaration", () => {
    const hostile = [
      "0; background:url(https://evil.example/x.png)",
      '0" onload="alert(1)',
      "0/*",
      "abc",
    ];
    for (const value of hostile) {
      const asDirection = flattenMarkup(
        renderRipple(true, "#fff", 80, value as unknown as number, 360, 3, "fixed")
      );
      expect(asDirection).toContain("--fp-ripple-direction:0");
      expect(asDirection).not.toContain("background");
      expect(asDirection).not.toContain("onload");

      const asWidth = flattenMarkup(
        renderRipple(true, "#fff", 80, 0, value as unknown as number, 3, "fixed")
      );
      expect(asWidth).toContain("--fp-ripple-width:360");
      expect(asWidth).not.toContain("background");
      expect(asWidth).not.toContain("onload");
    }
  });
```

Confirmed to fail with `cssNumber` removed and pass with it. (There's also a doubled blank line left after the block you added, just above `clamps the area name size` — worth tidying while you're in there.)

### 2. The two keys are missing from the config table

The prose is good, but `README.md:512` is the table a manual-YAML user actually reads, and it lists `rippleColor` and `rippleSize` without the new pair:

```
| `rippleDirection` | number   | `0`   | Direction the ripple arc is centred on, in degrees clockwise from the top. Only visible when `rippleWidth` is under `360`. Clamped to `0`–`360`. |
| `rippleWidth` | number       | `360` | Angular width of the ripple arc in degrees. `360` rings all the way round; narrow it for a sensor on a wall that cannot see behind itself. Clamped to `0`–`360`. |
```

The prose sentence beside it also runs to 227 characters where this file wraps at about 90 — worth rewrapping, and it can say *why* rather than only *what*:

> **Ripple width** and **Ripple direction** narrow the ring to an arc — for a sensor mounted on a wall, which cannot detect anything behind itself, so a ring that goes all the way round claims coverage it does not have.

---

## Still open from last time, not blocking

- **Every ripple now goes through the mask**, including default 360° ones, so existing plans get a compositing layer they did not have. Worth one look at a default ripple on a real plan before merge.
- **`direction` clamps rather than wraps.** `-90` on a compass means "left"; clamped it becomes "top". Wrapping (`((d % 360) + 360) % 360`) would be kinder, and is still safe. Your call.
- **CI has still not run on the head** — `checks` is empty on `7fa6e30`. A push should trigger it, which the fixes above will do anyway.

Apply those two and I'll approve straight away. Nice, quick work on the sanitizing.

jjj120:
I added the suggested test which passes.

I also changed the behavior of clamping/wrapping: The direction now wraps, since we want to be able to enter a direction of e.g. -90° and get 270°, not 0°. But the width still clamps. A width below 0° does not make sense and therefore gets clamped to 0°, and a width above 360° also just stays at 360°.

Also added the docs entries where they should go and added the info about wrapping and clamping.

And regarding the fact that every ripple now has the extra mask layer: I don't think that is a problem, since a mask that does not mask anything probably just gets ignored when rendered by the browser. So this should not make a big difference. I also don't know how I would solve this.

nicosandller (approved):
Approving. Everything is addressed, and the angle handling came back better than what I suggested.

```
npx tsc --noEmit    clean
npm test            1399 passed (17 files)
npm run build       clean
mergeable           yes
```

**The injection is closed.** Re-ran the probe from my first review across twelve cases, hostile and ordinary:

```
"0; background:url(https://evil.example/x.png)"  -> direction:0    width:360
'0" onload="alert(1)'                            -> direction:0    width:360
"0/*"  /  "abc"  /  null  /  undefined           -> defaults
-90                                              -> direction:270
450  /  360                                      -> direction:90  /  0
width 400  /  width -10                          -> width:360  /  width:0
170 / 90                                         -> unchanged
```

**Wrap for direction, clamp for width — and you were right to split them.** I suggested wrapping without saying what it should mean for width, and wrapping both would have been wrong: a width of `400` reading as `40` would turn "wider than a full circle" into a narrow sliver. Clamping it, and wrapping only the angle where `0` and `360` genuinely are the same point, is the correct reading. The two comments saying so are worth having.

**The adversarial test is in**, and I checked it earns its place under the new logic rather than assuming it still did: removing `cssNumber` fails it (`wrapAngle` alone yields `NaN`, not the default), and restoring it passes.

**Docs are in the right place.** `docs/configuration.md` is where that table lives now — you found the move I had missed when I quoted `README.md:512` at you. Both keys are there, with the wrap/clamp difference stated per key, and the prose is rewrapped.

---

## Closing out my non-blocking mask note

I said every ripple now goes through a conic mask including default 360° ones, and that it wanted a look. I checked rather than leaving it hanging:

```
CSS.supports('mask', 'conic-gradient(…)')          true
CSS.supports('-webkit-mask', 'conic-gradient(…)')  true
computed:  conic-gradient(from -180deg, rgb(0,0,0) 0deg, rgb(0,0,0) 360deg, rgba(0,0,0,0) 360deg)
```

At `width: 360` the transparent stop sits at exactly the end of the sweep, so the mask is opaque across the whole circle by construction — a default ripple is unchanged, and you kept the `-webkit-` prefix alongside. The only residual is a possible hairline at the seam angle, which no amount of reading settles; worth half a glance on a real plan, not a reason to hold this.

## Two trivia, merge without them if you like

- `render.ts:4851` — `return ((value % 360) + 360) % 360` has no trailing semicolon, unlike its neighbours.
- CI still has not run on this branch — `check-runs` is empty on `d2c4cea`, as it has been since `cafe4cd`. Worth a maintainer re-run before merge so this lands on a green build rather than an unrun one.

---

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ Threading, wrap, clamp, and the adversarial case — verified by mutation |
| 2 | Linked issue / intent | ✅ Solves #244, and does what was asked |
| 4 | README / config docs | ✅ Prose in the README, both keys in `docs/configuration.md` |
| 5 | Screenshots | ✅ Editor fields and the rendered arc |
| 6 | PR description | ✅ Brief, clear, links the issue |

Thanks for the quick, careful turnarounds on all three rounds — particularly for pushing back on the clamp/wrap question rather than just doing what the review said. Good first contribution.

PR #259, opened 2026-09-04
Replay: open means replay, closed means live

nicosandller:
## The bug

With `historyReplay.enabled: true` in the config, the plan could quietly stop being live — lights that were on drew off, and presence sensors with a ripple sat still even while tripping. Toggling that light from the plan really did switch it, because the service call is live either way, which made it read as a rendering bug rather than a replay one.

Replay could reach the plan through any path that *started* it, not only through someone opening the panel. Switching floors was one such path, and it parked the head at the **start** of the window — so the whole plan rendered from an hour ago, behind a panel that was shut and said nothing about it.

The previous fix (#256) narrowed the rule to "history, unless the head is parked at the end of the window". That is true but subtle, and it still left every not-parked-at-the-end path able to take over a live plan invisibly.

## The change

The panel being open is the whole question:

- **Closed → live.** `getRenderState()` returns `enabled: false` whenever the panel is shut, whatever replay has loaded and wherever the head sits. A plan nobody asked to rewind is indistinguishable from one with the feature off.
- **Opening starts replay** and parks the head at the *start* of the chosen window, so it begins where the calendar says it does.
- **Closing returns to live** — stops the clock, keeps the loaded events, so reopening on the same range costs no second history fetch.
- A floor switch or a window change reloads replay **only while the panel is open**.

That removes the need for a separate way back, so:

- The **Live** button is gone, and the corner button says `live` instead of `hide` (aria-label: "Close replay and return to live").
- The **Live/Replay chip** is gone — with the panel open there is only one mode left to be in.
- The header's **clock takes the accent** the chip was wearing, which answers the only question the panel still has: where in time is this. It matches the clock riding the timeline playhead.

## Tests

New tests assert the symptom on the rendered plan, not just in the controller: a ripple sensor that is tripping now keeps pulsing behind a closed panel with contradicting history loaded, shows the past once the panel opens, and goes back to now when it closes. Verified they fail against the old rule.

Rewritten: the chip tests (now about the highlighted timestamp), the "way back to now" tests (now about closing), and the floor-switch test (now asserts it does *not* reload while closed).

Full node suite and browser suite pass. The 3 node failures on this repo come from a stale nested worktree under `.claude/worktrees/` and are present on clean `main`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,11 +58,22 @@ and remain valid for backward compatibility.
 
 | Field | Type | Default | Description |
 | --- | --- | --- | --- |
-| `enabled` | boolean | `false` | Shows replay controls and enables loading history for mapped entities. |
+| `enabled` | boolean | `false` | Offers the **Replay** button on the card. Replay itself stays off until you press it. |
 | `lookbackSeconds` | number | `3600` | Initial replay window length in seconds. Must be positive. |
 | `defaultSpeed` | number | auto | Playback speed in simulated seconds per real second. `1` means real-time. |
 | `debug` | boolean | `false` | Log replay lifecycle (load, seek, play, pause) to the browser console. Off by default — seeking logs once per frame. |
 | `numericSteps` | number | unset | Thin numeric sensor drift to this many levels of each sensor's observed range, to cut the time a freshly loaded window takes to draw. Discrete states — lights, doors, covers — are never thinned. Unset keeps every point. |
+
+`enabled: true` puts a **Replay** button on the card and nothing else. The plan keeps
+showing live state until you press it; opening the panel loads the chosen window and
+puts the plan at the *start* of it, and the **live** button in the panel's corner
+closes replay and hands the plan back to Home Assistant. There is no separate on/off
+setting and no mode to be stuck in: open means replay, closed means live. The panel's
+clock — the highlighted timestamp in its header, and the one riding the timeline
+playhead — is where in time the plan is being drawn from.
+
+Closing keeps the window loaded, so reopening on the same range does not fetch it
+again.
 
 A busy plan can load thousands of points a numeric sensor never meaningfully moved
 through, and each one becomes a marker on the timeline. `numericSteps: 25` keeps the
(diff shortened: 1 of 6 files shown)


nicosandller:
Both comments are the same hole from two sides, so they get one answer.

`historyReplay.enabled` gates whether the panel is rendered at all, so switching it off with the panel open took the control off screen without taking replay with it — the plan kept drawing an hour ago with nothing anywhere to stop it. Opening had the mirror image: `toggleHistoryVisible(true)` would fetch history for a panel that was never going to be drawn.

Added `isReplayShowing()` — the config offers replay **and** the panel is open — and made all three paths read it:

- `getRenderState()` gates on it, so a plan with no panel on screen is live.
- `toggleHistoryVisible(true)` checks it before starting, so nothing loads for a panel that won't render.
- `setConfig` calls the new `syncToConfig()`, which shuts a panel the config no longer draws and stops the clock behind it. It replaces the old visibility check, so it still covers the ordinary closed-panel case.

Three tests, each verified to fail without the change: the plan goes back to live when the feature is switched off mid-replay, the clock stops and the panel shuts, and opening against a disabled config never calls `startReplay`.

One knock-on: the `card-badge` browser test mounted a card with no `historyReplay` block and set the open flag directly, which no longer amounts to replay showing. Its config now configures replay, which is what it was always modelling.

nicosandller:
Right, and it does bite: with the flag left set, `setConfig` on re-enabling would see `isReplayShowing()` as true, so `syncToConfig()` would leave the panel open — and it would come back open but *unstarted*, because `startReplay` had bailed on the way in.

`historyVisible` means "the panel is on screen", so both halves of that now decide it at once, before anything is written:

```ts
const open = visible && this._replayOffered();
this.state.historyVisible = open;
```

The disabled case falls into the same branch as an ordinary close — flag cleared, clock stopped — rather than into a state of its own. Pulled the config check out into `_replayOffered()` so `isReplayShowing()` and the toggle ask it the same way without one reading the flag it is about to set.

Two more tests, both verified failing without the change: opening against a disabled config leaves `isHistoryVisible()` false, and re-enabling the feature afterwards does not bring the panel back already open.

PR #271, opened 2026-09-06
Sunlight: a closed roller shutter is not a clear window (one rule for both light paths)

superflyer11:
A roller shutter bound as a window's own entity blocks a lamp's glow when closed, but currently still admits full sunlight. Windows default to glazed, and the sunlight path lacks the roll-motion exception already used by lamp glow.

This addresses the shared opening-transmission concern raised in #204. It extracts `openingGlassIsClear` and uses it in `glowClearFraction`, `glowClearSpan` and `openingSunFraction`:

```ts
return openingIsGlazed(o) && openingMotion(o) !== "roll";
```

The lamp behavior stays the same; sunlight now follows the roller shutter's open fraction. A closed roller shutter admits no sun, and a half-open shutter admits half its opening. The helper documents the existing blind/shade/curtain limitation: their inferred motion is `slide`, so distinguishing them from sliding glass still needs a separate covering flag. That follow-up is outside this fix.

## Demo and reproduction

The committed demo now enables `sunlight: true` and sets `motion: roll` on window `o1`, bound to `cover.living_room_window`. The demo cover has no shutter device class, so the explicit motion is necessary to exercise this rule. Its position control is already on the History view.

For an existing development dashboard, run `npm run ha:reseed`. Turn off the history generator, disable sun dimming, and pin the sunlight to `sunBearing: 0`. Move `cover.living_room_window` through 100%, 50% and 0%, returning to Plan each time. Full instructions are in `docker/README.md`.

## Screenshots from the development Home Assistant container

Home Assistant **2026.9.2**, Chromium 151, the same demo plan and sun bearing in every shot. Lamps and the history generator are off. The card is expanded to the full section width for readability. Look at the **centre opening in the top wall**; the other top window remains unchanged.

| Before: shutter closed, sunlight incorrectly passes | After: shutter closed, target patch disappears |
| --- | --- |
| Before: closed roller shutter still admits sunlight | After: closed roller shutter blocks sunlight |

| After: shutter fully open | After: shutter half open |
| --- | --- |
| After: open shutter admits the full patch | After: half-open shutter admits a narrower patch |

[Capture configuration, reproduction notes and measured SVG geometry](https://github.com/superflyer11/easy-floorplan/tree/4c70642da58b135b68f9cb7ebca1607184e49a61/assets/pr271-review). The cover's actual Home Assistant position was checked before each capture. At 100% the target beam starts 140 canvas units wide; at 50% it is 70 units; at 0% it is absent. The old build still draws a 140-unit opening at 0%.

## Validation

On `7537d2f`, incorporating current `main` (`3f89501`):

| Check | Result |
| --- | --- |
| `npm run typecheck` | Passed |
| `npm test` | 1,521 passed, 20 files |
| `npm run test:browser` | 55 passed |
| `npm run build` | Passed |

The existing regression cases cover the roller-shutter divergence, sun/glow agreement across covering device classes, and the helper truth table including the documented blind/shade/curtain gap. The maintainer verified that reverting the helper makes the relevant tests fail. No browser page errors occurred during the real Home Assistant captures.

The merge preserves upstream's new `awning` motion while retaining the helper's narrower input type. No new configuration fields or migrations are introduced by the fix; `docs/lighting.md` documents its behavior and limitation.


nicosandller (changes requested):
This is a good fix and the write-up made it fast to check. I verified the claim rather than reading it — the divergence is exactly as you describe, and the helper closes it:

```
device_class   motion   CLOSED sun          CLOSED glow
shutter        roll     1  ->  0            0            <- the bug, and the fix
blind          slide    1      1            1            <- the gap you documented
shade          slide    1      1            1
curtain        slide    1      1            1
window         -        1      1            1
```

Left column is `main`, right is this branch. Sunlight gains the rule; nothing else moves.

The three tests earn their place — I reverted `openingGlassIsClear` to plain `openingIsGlazed` and three fail, one of them an existing `glowClearFraction` test, which is a nice demonstration that the helper is now load-bearing for both paths rather than a wrapper.

Documenting the blind/shade/curtain gap in the helper and pinning it with a test, instead of leaving the old docstring claiming coverage it never had, is the right call. The docstring was promising `blind, shutter, shade, curtain or awning` and delivering `shutter` alone.

```
npx tsc --noEmit    clean
npm test            1467 pass (18 files)
npm run test:browser   24 pass
npm run build       clean
```

Two things before I approve, both about being able to *see* it rather than about the code.

---

## 1. It can't be tested on the dev container as it stands

I went to check this on the demo plan and could not make it happen. Two reasons, both in `docker/config/floorplan-demo.yaml`:

**No opening reaches the new rule.** The fix needs `motion: roll` *and* glazing. The demo has exactly one roll-motion opening, `o3`, and it is a **door** — not glazed, so it was already stopping light through the aperture rule and behaves identically before and after. The only cover-bound window, `o1`, takes its motion from its entity's device class:

```
cover.living_room_window   device_class = None
cover.hall_window          device_class = None
cover.garage_door          device_class = garage
```

None is `shutter`, so `openingFromDeviceClass` gives `motion: undefined`, and `o1` reads as clear glass before and after.

**The sunlight layer is off.** The plan sets `sunDimming: true` but not `sunlight: true`, so the beam-through-openings layer this PR changes is not drawn at all.

Minimal change that makes it visible — `sunlight: true` on the plan, and one line on `o1`:

```yaml
- id: o1
  type: window
  x: 500
  y: 100
  length: 140
  angle: 0
  motion: roll          # a roller shutter, which is what binding this cover means
  entity: cover.living_room_window
```

I confirmed that recipe gives `closed sun=0, half-open sun=0.5` on this branch against `sun=1` at every position on `main` — so it shows the fix, and `cover.living_room_window` already has a position slider in the demo to drive it.

## 2. No screenshots

This changes what the plan looks like — a patch of sunlight that used to fall through a closed shutter and now doesn't. That is the kind of change a still says more about than a table of fractions, and it is the one thing here nobody can check by reading.

With the demo change above it should be two shots of the same plan at the same sun position, shutter up and shutter down.

---

## Checklist

| # | Item | |
| --- | --- | --- |
| 1 | Tests | ✅ Three, verified by mutation; the agreement table across every covering class is the right shape |
| 2 | Linked issue / intent | ⚠️ No issue linked — it comes from the fourth point on #204, which the description explains. Non-blocking, worth a line for the record |
| 4 | Documentation | ✅ `docs/lighting.md` says what changed and names the gap |
| 5 | Screenshots | ❌ See 2 |
| 6 | PR description | ✅ Among the clearest on this repo — the before/after fractions are what let me verify it in one pass |

Happy to approve once it can be seen on the demo plan and there are stills. And yes to the covering-flag follow-up — `ambientOpeningTransmission` on #204 is the third copy of this rule, so a shared helper it can call is worth having before that lands.

superflyer11:
Addressed both blockers in 7537d2f and the updated PR description:

- The committed demo now sets `sunlight: true` and `motion: roll` on `o1`. `docker/README.md` explains reseeding, stopping the generator, fixing the sun bearing, and testing the existing cover position control.
- The description contains real development-container screenshots at 100%, 50% and 0%, plus the old build at 0% for comparison. All use the same demo plan and sun bearing. The target sunlight opening measures 140 units, 70 units and absent respectively; the old build still admits the full 140 units when closed. The capture configuration and measurements are linked alongside the images.

Also merged current `main`, resolving the `openingMotion` conflict while preserving `awning`. Type checking, 1,521 unit tests, 55 browser tests and the build pass locally; GitHub's build and browser checks are green. The description explicitly records #204 as the source of the intent. The broader covering-flag work remains a separate follow-up.

Ready for another review.

nicosandller (approved):

Issue #90, opened 2026-07-30
Furniture editor - new furniture request


gorischek:
Hi,
For example, I have a wardrobe with seven doors in my bedroom. Currently, this combination is difficult to represent. Therefore, I would like an option to define whether a wardrobe has one or two doors.
Br, Michael

nicosandller:
Thanks for the feedback @gorischek !

Do you have an idea on how we could implement it in the UI? Currently furniture is kinda fixed and predefined. Maybe the best approach is to have an open library where people can draw furniture and share it with everyone?

This would be ambitious but it would be a nice long term collaborative way to approach this I think.

gorischek:
Yes, I also think that would be the best way. Otherwise, you'll just keep getting more and more requests. This way, everyone could extend it (in the true spirit of open source), and you'd "only" have to maintain the framework.

It would be great if you could make the existing objects available as templates, so people could use them.

PKeller34:
Maybe some kind of "Import" feature for symbols could be implemented?   Then folks could draw any symbol they like and use it.  I'm a total NOOB with Git, but that might be a way to allow people to contribute new symbols for anyone to download and use.  It would take some work to create a "Symbol design guide" for Easy Floorplan so that symbols have the correct attributes to work in Easy Floorplan, but that would probably pay off very quickly since you wouldn't have to spend time on 1000 symbol requests.

I believe you mentioned the symbols are actually SVG.   As I understand it, the SVG format allows for parameterized symbols so the glyph can change in real time or even animate based on parameters.  If the Symbol Design Guide explains what parameters Easy Floorplan can utilize then other requests like [https://github.com/nicosandller/easy-floorplan/issues/54](https://github.com/nicosandller/easy-floorplan/issues/54) or maybe [https://github.com/nicosandller/easy-floorplan/issues/8](https://github.com/nicosandller/easy-floorplan/issues/8) could be handled by users themselves.

superflyer11:
Two separate things in this thread, so I had a go at the smaller one and want to
float a design for the bigger one.

**The original ask** — a wardrobe with more than two doors — is now #117. `doors`
(1–12, default 2) divides the front into panels; the default renders the existing
glyph exactly, so no current plan moves. @gorischek, that covers the seven-door case.

**The library.** I agree with @PKeller34 that user-supplied symbols is the way to
stop the request queue, but I'd argue for *not* accepting SVG markup, and here's the
tradeoff.

[...]

PR #273, opened 2026-09-07
Add hot tub and shower stall for use in corners

larsalthof:
Both hot tub and shower stall are quarter circles to be used in corners. They can be stretched, but are not supposed to be


nicosandller (approved):
Two good additions — corner pieces are a real gap in the bath category, and both follow the geometry format exactly. I've verified them in a browser at their default size, stretched 2:1, and at 45px, and pushed the fixes below to your branch.

## The blocker, and why it wasn't your fault

`npm test` failed on this branch:

```
FAIL  src/symbols.test.ts > the shipped furniture library > answers to every name in the FurnitureType union
+   "cornerShowerCurved"
+   "quarterTub"
```

`symbols.test.ts` asserts that the shipped library and the `FurnitureType` union in `src/types.ts` are the **same set**, in both directions — so two JSON files with no union entry fail it.

You had no way to know that: `furniture/README.md` said adding a symbol needs *"no code, no build step, no entry in a list somewhere else."* There is one, so I've corrected the README as part of this PR rather than leaving the next contributor to hit the same wall.

CI never told either of us: this PR had **no checks at all**, because a first-time contributor's workflow run needs maintainer approval. I've approved it — it's green now.

## What I pushed (3 commits, disclosed)

- Added both ids to `FurnitureType` and to the union list in `symbols.test.ts` — suite green, 1464 pass.
- Fixed `furniture/README.md` so it names the list.
- Lowercased both `name`s. Every other symbol in the picker is lowercase (`hot tub`, `round table`, `sectional (L)`), and Title Case stood out in the list.
- Added the keywords these pieces are actually searched by: `corner`, `stall`, `cubicle` on the shower, `corner`, `quarter`, `spa`, `whirlpool` on the tub. Kept `dusche`.
- Trailing newlines, and the README's "26 gray line-art diagrams" → 28.

## Two notes on the drawings — your call, not blocking

Rendered next to the existing `hotTub` for scale:

- **`cornerShowerCurved` has no interior detail at all** — it draws as a plain quarter-round with nothing inside it, so at picker size it isn't identifiable as a shower rather than any other corner shape. Every other bath symbol has something: `bathtub` a drain, `toilet` a seat, `vanity` a basin, `hotTub` four jets. One drain would do it:

  ```json
  { "circle": [50, 50, 4], "role": "line" }
  ```

- **`quarterTub`'s inner rim pinches to a point** at the corner: it ends at `95,95` while the body's corner is at `100,100`, and the `detail` sweep converges on the same spot. At size it reads as a leaf rather than a tub. Pulling the rim's end back (`L 95, 88` and a short curve into the corner) opens it up.

Both stretch acceptably, and you're right that they're not meant to be — the `size` defaults keep them square when placed.

No linked issue; #90 is the standing "new furniture request" one if you want to reference it. Non-blocking, just noting it.

Approving. Thanks for these — corner fixtures are genuinely useful and the geometry is clean.

PR #274, opened 2026-09-08
Point TypeScript at the files nobody was checking

### LINKED ISSUE #246: Root config files are not type-checked
Split out of the review on #245, where this bit for real.

`tsconfig.json` has `"include": ["src"]`, so nothing type-checks the root config files — `vite.config.ts`, `vitest.browser.config.ts`, `vitest.setup.ts`, `docker/*.mjs`. `npm run typecheck` is clean whatever they contain.

## How it showed up

#212 added a `test:` block to `vite.config.ts`; #245 added one too, in a different place. Git merged both without a conflict and left a duplicate object key:

```ts
export default defineConfig({
  test: { setupFiles: ["./vitest.setup.ts"] },   // #212
  build: { … },
  test: { exclude: [...] },                      // #245 — silently wins
});
```

The second key wins, `setupFiles` is dropped, and with it the `PointerEvent` polyfill that `src/replay-history/history-replay.test.ts` depends on. The symptom is two failures in a file neither PR touched, which is a misleading place to start debugging. `tsc` says nothing, because it never looks at the file. `npm test` does catch it, so CI would not let it through — but only after a detour.

TypeScript reports duplicate object keys (`ts(1117)`) when it checks the file at all. It just never checks this one.

## What would close it

Add the root config files to the type-check. Either widen `include`, or keep the app's `tsconfig.json` narrow and add a `tsconfig.node.json` covering the config files, with `npm run typecheck` running both. A typo'd Vite or Vitest option is the same class of silent failure, so the value is not only the duplicate key.

nicosandller:
Closes #246.

`tsconfig.json` includes only `src`, so nothing type-checked the build and test configuration or the docker scripts. `npm run typecheck` was clean whatever they contained.

## The bug from the issue, reproduced and then caught

Putting the #212 + #245 merge back into `vite.config.ts`:

```ts
export default defineConfig({
  test: { setupFiles: ["./vitest.setup.ts"], … },   // #212
  test: { exclude: [...] },                         // #245 — silently wins
  build: { … },
});
```

**Before** — `npx tsc --noEmit`:

```
✓ clean
```

The duplicate key is invisible, `setupFiles` is dropped, and with it the `PointerEvent` polyfill `src/replay-history/history-replay.test.ts` depends on.

**After** — `npm run typecheck`:

```
> tsc --noEmit && tsc -p tsconfig.node.json

vite.config.ts(10,3): error TS1117: An object literal cannot have multiple
                      properties with the same name.

exit code 2
```

Same repo, same file, same TypeScript. The only difference is that something now looks at it. CI already runs `npm run typecheck`, so this protects immediately.

## Why a second project, not a wider `include`

These files run under Node and import Node builtins; `src` is browser code bundled by Vite. Merging them would hand `src` the Node globals it must never reach for. So `tsconfig.node.json` covers `vite.config.ts`, `vitest.browser.config.ts`, `vitest.setup.ts` and `docker/*.mjs`, and `npm run typecheck` runs both.

`checkJs` is on for the `.mjs` scripts with `noImplicitAny` **off**. They walk untyped JSON out of Home Assistant's `.storage`, and demanding an annotation on every parameter would be a rewrite rather than a fix. Everything else in `strict` stays on — which is what catches the real mistakes.

## The thirteen it surfaced

All genuine, all fixed here:

- **`never[]` inference.** `data: { items: [] }` in `prepare.mjs` infers as `never[]`, so every push into the resource and dashboard stores is an error. One `@typedef` fixes both.
- **`yaml.load` returns `unknown`** — the seed's shape is now stated rather than assumed.
- **The catch variable is `{}` under strict**, so `error?.code` needed narrowing.
- **`sun-at.mjs` subtracted two `Date`s**, which JavaScript allows and TypeScript does not. Now `getTime()` on both sides.

The Date change is the only one that touches runtime behaviour, so I checked it rather than assuming: the two forms produce an identical day-of-year, and `node docker/sun-at.mjs 15.5` still prints what it did.

## Test

`src/tsconfig-coverage.test.ts` walks the repo and asserts the two projects' `include` lists between them reach every checkable file — so adding a `playwright.config.ts` that nobody checks fails at the point it is added, rather than the next time a merge quietly drops half of one. It also asserts `npm run typecheck` actually invokes the second project, since covering a file in a config no script runs checks nothing.

Includes a test that the coverage check can still say *no*, because the main assertion passes just as happily if the matcher breaks open.

`1468` tests passing, `npm run typecheck` and `npm run build` clean.

## Docs

CONTRIBUTING gains a "Two TypeScript projects" section next to the existing "Two test suites" — which project a new file lands in is now the other thing that is easy to get wrong.

---

**Note on the screenshot:** this is a tooling change with no UI, so the artifact is the terminal transcript above. GitHub's `user-attachments` CDN only accepts uploads through the web UI, so I can't attach the rendered image from here — I've sent the PNG to you in the session if you'd like to drag it in.


src/tsconfig-coverage.test.ts
```diff
@@ -0,0 +1,117 @@
...
+ * actually reach every one of them. Adding a `playwright.config.ts` or a
+ * second docker script that nobody checks fails here, at the point it is
+ * added, rather than the next time a merge quietly drops half of one.
+ */
+import { describe, it, expect } from "vitest";
+import { execFileSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+import { join, resolve } from "node:path";
+
+const ROOT = resolve(__dirname, "..");
```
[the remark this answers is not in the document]
nicosandller: Fixed in 17ee248.

Worth saying why it wasn't already failing: it resolved because vitest's transform injects a CJS shim, so the test passed and nothing pointed at it. That's a bad thing for this file in particular to be leaning on — it's the repo's statement about what is and isn't type-checked, so it shouldn't be the one file that needs a shim to load.

Now derived from `import.meta.url`. I checked the expression resolves to the repo root under plain `node` as well, not just under the test runner.

src/tsconfig-coverage.test.ts
```diff
@@ -0,0 +1,147 @@
...
+ * — threw at import time, so a harmless edit to either tsconfig took the whole
+ * suite down while `tsc -p` still accepted the file. Reproduced before
+ * changing it.
+ *
+ * `readConfigFile` is the compiler's own jsonc reader and
+ * `parseJsonConfigFileContent` its own `include`/`exclude` resolution, so what
+ * comes back is the file list `tsc` will really check. That retires the glob
+ * matcher along with the parser, and it means a config form nobody here
+ * anticipated — `files`, `extends`, a negated `exclude` — is understood for
+ * free rather than silently mis-read.
```
nicosandller, line 81: Right, and I reproduced it before changing anything — a block comment and a trailing comma added to `tsconfig.node.json`, both legal jsonc:

```
npx tsc -p tsconfig.node.json --noEmit   →  accepts it
npx vitest run src/tsconfig-coverage.test.ts
  SyntaxError: Expected property name or '}' in JSON at position 18
  Tests  no tests
```

A harmless edit to either tsconfig took the whole suite down while `tsc` still compiled the project. Bad anywhere; worse in this file, which exists to be the repo's statement about what is checked.

Fixed in 15e4b6d, and further than the comment asks. Rather than write a sturdier stripper I handed the whole question to the compiler: `ts.readConfigFile` is its own jsonc reader and `ts.parseJsonConfigFileContent` its own include/exclude resolution, so what comes back is the file list `tsc` will really check.

That retires **both** guesses — the parser and the hand-rolled glob matcher underneath it, which was the same class of bug waiting to happen. A config form nobody here anticipated (`files`, `extends`, a negated `exclude`) is now understood for free rather than silently mis-read, and a tsconfig that cannot be parsed throws by name instead of resolving to an empty list and blaming the other project.

One thing this exposed on the way: the negative control was fake. It asked whether the matcher rejected a path for a file that does not exist — which passes just as happily when the matcher is broken open. It now writes an actual `playwright.config.ts` at the root, and the suite fails naming it. I also swapped in a sharper assertion, that each file lands in the project that should *own* it: plain coverage is satisfied by one project swallowing everything, which would quietly hand `src` the Node globals the split exists to keep away from it.

package.json
```diff
@@ -38,9 +38,10 @@
     "lit": "^3.1.0"
   },
   "devDependencies": {
+    "@types/node": "^24.13.3",
     "@vitest/browser": "2.1.9",
```
nicosandller, line 42: Correct, and my reason for `^20` was a bad one: it was the Node on my machine, not the Node this project runs. Every workflow is on 24.

Not cosmetic either. Reproduced with a docker script using `fs/promises.glob`, which exists on CI's Node 24:

```
docker/probe.mjs(2,10): error TS2305:
  Module '"node:fs/promises"' has no exported member 'glob'
```

So the type-check this PR adds could have rejected a script for using an API that is right there at runtime. The reverse drifts more quietly still — newer typings would describe APIs an older CI has never had.

Bumped to `^24` in 5c1c3f7. No new errors surfaced in the config files or the docker scripts.

The more useful half of your comment was the second one, about the two being able to disagree at all. Version alignment agrees once and drifts later, so I asserted it instead of documenting it: the coverage test reads `node-version` off the workflows and compares it with the `@types/node` major. Verified in all three directions —

- typings behind CI → `expected '20' to be '24'`
- both workflows bumped ahead → `expected '24' to be '26'`
- the two workflows disagreeing with *each other* → `expected [ '26', '24' ] to have a length of 1`

That last one falls out of requiring a single answer, which is what makes the comparison mean anything.

One thing I deliberately did **not** add: an `engines` field. It would pin this down further, but which Node majors the package supports is a policy decision about consumers rather than a consequence of type-checking the build config — that's yours to make, and easy to add on top if you want it.

PR #278, opened 2026-09-08
Undo survives card-mod (#257)

### LINKED ISSUE #257: Custom card_mod disables undo function
When I add `card_mod` to my floor plan, the undo function in the visual editor stops working. The button is greyed out and the Ctrl+Z shortcut does nothing. As soon as I remove it, everything works fine.

My redacted floor plan:

```yaml
width: 1080
height: 1480
grid: 7
overlayScale: plan
grid_options:
  columns: full
floors:
  ...[omitted]...
showDeadSpaces: true
sunlight: true
north: 270
sunReach: 0.3
symbols:
   ...[omitted]...
card_mod:
  style: |
    ha-card {
      max-height: 1000px;
    }
```

Easy Floorplan: v1.7.2
Home Assistant - Core: 2026.8.1
Home Assistant - Frontend: 20260729.6

nicosandller:
Fixes #257.

## Cause

Not a guess — this is card-mod's own [`src/patch/hui-card-element-editor.ts`](https://github.com/thomasloven/lovelace-card-mod/blob/master/src/patch/hui-card-element-editor.ts). It patches the card editor on **both** halves of the round trip:

```js
setConfig(_orig, config, ...rest) {
  const newConfig = JSON.parse(JSON.stringify(config));  // identity gone
  this._cardModData = { card: newConfig.card_mod, entities: [] };
  delete newConfig.card_mod;                             // stripped going in
  _orig(newConfig, ...rest);
}

_handleUIConfigChanged(_orig, ev, ...rest) {
  ev.detail.config.card_mod = cmData.card;               // added coming out
  _orig(ev, ...rest);
}
```

`ev.detail.config` is the object the editor emits — and the same object it keeps as `_lastEmitted`. card-mod writes `card_mod` onto it *after* we've stored it.

So on the next `setConfig`:

- the deep clone means `config !== this._lastEmitted`, so the identity fast path never fires;
- the incoming config has `card_mod` **stripped**, while our stored `_lastEmitted` has since **grown** it;
- `configsEqual` therefore says "different", every edit reads as an external YAML change, and `this._history = []` runs on each one.

Undo button permanently disabled, Ctrl+Z inert — exactly as reported. Without card-mod the identity check passes and nobody ever notices.

## Fix

One line: `_lastEmitted` is a `structuredClone` of what we emit. Holding a reference to an object handed out on an event and comparing against it later was the mistake — nobody else can write to a copy.

It also lands the right semantics for card-mod users: the editor never sees `card_mod`, so it's never in an undo snapshot, so undo can't revert a card-mod edit made in the YAML tab.

## Tests

`src/editor-undo-card-mod.test.ts` reproduces card-mod's patch from its own source rather than describing it, so it fails for the reason the user's dashboard did. Verified against the unfixed editor:

| test | without the fix |
| --- | --- |
| keeps one undo step per edit | `expected +0 to be 1` — the greyed-out button |
| undo actually walks back | `expected 1200 to be 1080` — Ctrl+Z inert |
| still drops the stack for a genuine external edit | fails (guard's real job, kept) |
| behaves the same without card-mod | passes, as it always did |


nicosandller:
Both applied.

**The guard** — you're right that mine was narrower than card-mod's, and it's a real difference, not just wording. Its handler tests the saved *record*:

```js
const cmData = this._configElement?._cardModData;
if (cmData) {
  ev.detail.config.card_mod = cmData.card;
}
```

`_cardModData` is created by its `setConfig` wrapper whether or not the config had a `card_mod`, so once a config has been through, it writes the key back unconditionally — `undefined` included. The test now models `_cardModData` itself and guards on that.

Worth noting that branch is benign either way: `JSON.stringify` drops an `undefined` value on the way back in, so `configsEqual` compares `undefined` to `undefined` and agrees. The breakage needs an actual `card_mod` block, which is what the reporter had.

**The cleanup** — added `afterEach(() => { document.body.innerHTML = ""; })`. The editor binds `keydown` and `focusin` on `window` in `connectedCallback`, so an editor left mounted keeps listening through every test after it.

Re-verified against the unfixed editor after both changes: same three failures, same reasons (`expected +0 to be 1`, `expected 1200 to be 1080`).

PR #297, opened 2026-09-12
Furniture can do what a room can: tap, hold and double-tap actions

### LINKED ISSUE #284: Behavior for Furniture
Current the behaviour on all furniture, (including stairs), onl allows you to go to a floor:

Image

I would like the option to select either a floor or the tap actions like we have for areas:

Image

nicosandller:
Closes #284.

> Currently the behaviour on all furniture, (including stairs), only allows you to go to a floor … I would like the option to select either a floor or the tap actions like we have for areas.

The editor's Behavior group for a staircase

The real editor, a staircase selected. **Behavior** now carries the three actions under *Go to floor* — the panel from the issue's second screenshot. (Home Assistant's own form components don't exist outside HA, so each row is drawn by a stand-in from the schema the editor emits: the labels, order, helper text and control types are the editor's, the styling is not HA's.)

```yaml
furniture:
  - id: tv
    type: tv
    entity: media_player.living
    tap_action: { action: toggle }
    hold_action: { action: more-info }
```

## How it sits with `goToFloor`

`goToFloor` is to a piece what the zoom is to a room: what a tap does when nothing else is set. So the two **compose rather than compete**:

| | tap | hold / double-tap |
|---|---|---|
| `goToFloor` alone | changes floor, as before | — |
| `tap_action` set | the action | free |
| `goToFloor` + `hold_action` | changes floor | the action |
| `tap_action: { action: none }` | nothing | free |

That last row is the point of reading `none` as *configured* rather than absent — it's the only way to say "this staircase should not move me".

`furnitureActionForGesture` is deliberately a mirror of `areaActionForGesture`, and a test asserts the two agree on identical input, since "like we have for areas" is the requirement rather than an analogy.

## Three things beyond the routing

- **An inert piece stays inert** — no button role, no tab stop, no listeners. A gray diagram that announces itself as a button and then does nothing is worse than one that says nothing.
- **Hold and double-tap are declared to `actionHandler` only when configured**, so a floor-change-only staircase still answers a tap immediately rather than sitting out both timers first.
- **The tooltip stops promising a floor it will no longer go to.** It's the only thing on screen saying what a tap does, so leaving it reading "Go to Upstairs" would be a lie on exactly the plans this was asked for.

## Tests

- **7 node tests** on the resolver: absence, per-gesture isolation, entity fallback, an action naming its own target, the navigate case with no entity anywhere, `none` as configured, and agreement with the area resolver.
- **8 browser tests** on the mounted card, since none of the three points above is visible to a test of the resolver — including the two that matter most: a configured tap does **not** change floor, and a hold action leaves the tap alone.

One fixture note worth flagging: every mount gets fresh floor ids, because the card remembers the last floor you were on in a module-level map that outlives any one card. Reusing ids made a later test mount on the floor an earlier one had navigated to, which looked exactly like the feature failing.

`1519` node + `63` browser passing; typecheck and build clean.

## Docs

New "Actions on furniture" section in `docs/behavior.md` with the table above and the figure, the furniture block in `docs/configuration.md`, and a README bullet.

## One thing I noticed and did not fix here

The panel-group coverage test (`src/editor-forms.test.ts`) restates the editor's group tables in its own constants rather than reading them, so the two can drift — you could add a field, list it in the test, forget `editor.ts`, and the suite stays green while the field renders nowhere. It caught my omission this time, but only against its own copy. Left alone to keep this PR to the issue; worth its own change.


nicosandller:
All four addressed in `8bfe18f`. Two were real bugs rather than polish.

**A no-op counted as a gesture.** I used `!!furnitureActionForGesture(...)` where the rooms directly above use `hasAction(...?.config)` — and `{ action: "none" }` returns a record, so it read as a live gesture. Three consequences, all of which contradict the comment sitting right there about inert pieces staying inert:

- a piece whose only action was `none` became a button with a tab stop that does nothing when you press it;
- a staircase with `goToFloor` **plus** `tap_action: none` kept its hit target with its tap explicitly switched off;
- `hold_action: none` / `double_tap_action: none` spent the handler's timers, so every tap waited out a hold that could never fire.

These ask `hasAction` now. The one thing I kept separate is "was a tap configured at all" — `none` included — because that is what suppresses the floor fallback: `tap_action: none` is how a plan says "draw the stairs, don't let them navigate", and `_onFurnitureAction` already decided it that way. So the *suppression* still honours a no-op tap; only the *interactivity* stopped counting it.

**Action-only furniture was an unnamed button.** `renderFurniture` contributes paths and nothing else, and the floor tooltip that names the other case is missing precisely when there is no floor to go to. It gets an `aria-label` now — the entity's friendly name, else the entity id, else the symbol it's drawn as, humanised (`coffee-table` → "coffee table") — and only where the `<title>` isn't already naming it, since an aria-label would override it and "stairs" is a worse name than "Go to Upstairs".

**The hold/double-tap coverage gap was exactly as described.** The browser case dispatched `tap` on a piece configured with `hold_action`, so it proved the floor fallback survived and nothing more. Covered now at both observable exits — the `hass-more-info` event and the service call — and I confirmed it bites: making `_onFurnitureAction` drop non-tap gestures fails both new tests, and passed every old one.

Plus the missing noun in the `editor-forms.ts` comment.

I checked each new test fails against the old code before trusting it: reverting `hasAction` fails the two no-op tests, removing the `aria-label` fails the two naming tests.

**1526 node + 69 browser** passing, typecheck and build clean. `docs/behavior.md` gained the `none`-is-inert rule and the naming behaviour. No screenshot on this one — nothing here changes what the plan looks like; the visible surface is unchanged and the new behaviour is a missing tab stop and an accessible name.

nicosandller:
All five addressed in `4e55ea4`. The central one was right, and it was the same mistake in two places.

**"Configured" is not "able to run".** `hasAction` only answers "present and not `none`", but `executeAction`'s own guards go further — a `more-info` with no entity, a `navigate` with no path, a `call-service` with no service all pass `hasAction` and then do nothing. So a piece configured that way drew a button with a tab stop answering to nothing, and armed the hold and double-tap timers for gestures that could not fire, which makes **every tap wait out a hold that was never going to happen**. The codebase already had the right check — `gestureDoesSomething`, which `itemIsInteractive` has always used — so I exported it rather than writing a second one.

The accessible name asked the same weak question, which is the inline comment's point: an unusable tap could beat a working hold. It now takes the first gesture that could run *and* acts on an entity.

**On the `navigate`/`url` half of that comment, I went partway and want to flag where I stopped.** Those actions reach for a path, not a subject, so they no longer *supply* a name. But I kept the piece's own `entity` as the fallback underneath, which the comment reads as also wrong. My reasoning: `Furniture.entity` is what the drawing is bound to — the plant's soil sensor, the shelf's light — so it names the *thing*, independently of what any gesture does. For a button whose action has no subject of its own, "Shelf light" beats "table". Happy to drop the fallback if you read it the other way.

**The name could also go stale**, which I'd have missed. It reads live state, but a hass tick only re-renders when a watched entity moves — and an action's target deliberately isn't watched, since nothing about it is drawn. Rather than widening `collectWatchedEntities` (which would have a replay fetching history for an entity the plan never shows — the cost the comment itself names), there's a second set: consulted for whether to re-render, absent from what gets drawn. Without it a button kept announcing whatever it found first, including the raw id for an entity that didn't exist yet.

**`furnitureForm`'s fields had no direct test** — correct, the grouping test only compares names against a duplicated list. Six now cover presence, the `none` default, stored actions round-tripping, the `goToFloor` patch not eating them, and the tap helper appearing only on a staircase.

**The stranded JSDoc was real** — my furniture block had been inserted between `areaActionForGesture`'s comment and its declaration. Moved below it. Second one of these across the two PRs; I'm now checking what sits above an insertion point.

Each behavioural fix verified to fail against the old code first.

**1538 node + 73 browser** passing, typecheck and build clean. `docs/behavior.md` updated for both rules.

One note for whoever merges: the rooms at `floorplan-card.ts:1142` still use `hasAction` for their timers and have the same latent issue. I left it alone as pre-existing and out of scope here — say the word and I'll do it, in this PR or its own.

nicosandller:
Right, and it's a defect this PR introduced. Fixed in `15e8d39`.

I traced it rather than taking it on faith, because the claim depends on `actionHandler`'s internals: `handleKeyDown` accepts Enter/Space and calls `end(ev)`; `end` sees no prior pointer press, so `this.held` is false and the hold branch is skipped; with no `click` type to inspect it falls through to `fireActionEvent(target, "tap")`. Keyboard activation is always a tap. And for a hold-only piece the tap resolves to no action, with no floor to fall back to — so `_onFurnitureAction` returns having done nothing. Focusable, announced as a button, inert when pressed.

**Of the two remedies offered I took the second**, and the reasoning matters more than the diff:

Giving the keyboard a path to the hold would mean treating its activation as "the first gesture that does something". That requires telling a keyboard activation from a pointer tap, which the `action` event doesn't carry — and applying it blind would make a *pointer* tap fire the hold action, contradicting the config that deliberately put it on hold. I could add a competing `keydown` listener, but it would race `actionHandler`'s own on the same element and need `stopImmediatePropagation` to win, which is a fragile thing to rely on.

So the role and tab stop now come from the tap specifically — a configured tap that can run, or the floor change standing in for it. A hold-only piece keeps its listeners and still answers a pointer hold; it just stops claiming to be a control nobody can operate, and loses its accessible name with them.

Worth being explicit about what that leaves: **there is no keyboard path to a hold action on furniture.** That isn't a regression, it's the existing state of the whole card — hold and double-tap are pointer gestures for items and rooms too, because the keyboard has one activation and they are the second and third gestures on it. If you'd rather close that gap properly it's its own piece of work across all three element kinds, not something to bolt onto furniture here.

One earlier test had to change with this: I'd asserted a hold-only piece gets `role="button"`, which was the bug written down as an expectation. It now asserts the opposite, plus that the hold still reaches `executeAction` — and I confirmed it fails if the role is handed back to every gesture.

**1538 node + 74 browser** passing, typecheck and build clean. `docs/behavior.md` states the rule.

nicosandller:
Fixed in `dac9eaa`, though not quite the way the comment suggested — the catalogue turned out to be a better source than any regex.

The point stands: the fallback read the raw symbol id, and those ids are written for configs rather than for reading aloud, so a screen reader announced `fishTank` verbatim. But splitting camelCase would only get you "corner shower curved", and `furniture/cornerShowerCurved.json` already says `"name": "curved corner shower"`. Every built-in carries one, and so does any symbol a config defines itself. So it reads `findSymbol(catalog, f.type)?.name` now, which meant threading the card's resolved catalogue through to the resolver.

Unpicking the id is what's left for a type nothing in the catalogue answers to — a symbol since removed, or a typo — and that path does now split camelCase, as suggested.

**One thing I'd have shipped unnoticed without checking:** my browser test for the symbol fallback used a *built-in*, which the resolver finds through its own default catalogue whether or not the card passes one. It asserted the name and proved nothing about the wiring — reverting the card's argument left it green. There's a config-defined symbol beside it now, which does fail without it.

**1587 node + 75 browser** passing (the count moved because this branch also picked up your `main` merge with ambient daylight), typecheck and build clean.

